PR #16, opened 2015-04-06
Add ability to stop streaming

jacklund:
This adds the ability to stop and restart streaming. Stopping is accomplished by calling `KafkaEx.stop_streaming`, and restarted by calling `KafkaEx.stream` again.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -201,7 +201,7 @@ defmodule KafkaEx do
 
     {:ok, pid}  = GenEvent.start_link
     :ok         = GenEvent.add_handler(pid, handler, [])
-    send(worker_name, {:start_streaming, topic, partition, offset, pid, handler})
+    send(worker_name, {:stream, topic, partition, offset, pid, handler, 0})
     GenEvent.stream(pid)
   end
 
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -199,10 +199,15 @@ defmodule KafkaEx do
     offset      = Keyword.get(opts, :offset, 0)
     handler     = Keyword.get(opts, :handler, KafkaExHandler)
 
-    {:ok, pid}  = GenEvent.start_link
-    :ok         = GenEvent.add_handler(pid, handler, [])
-    send(worker_name, {:stream, topic, partition, offset, pid, handler, 0})
-    GenEvent.stream(pid)
+    stream      = GenServer.call(worker_name, {:create_stream, handler})
+    send(worker_name, {:start_streaming, topic, partition, offset, handler})
+    stream
+  end
+
+  @spec stop_streaming(Keyword.t) :: :ok
+  def stop_streaming(opts \\ []) do
+    worker_name = Keyword.get(opts, :worker_name, KafkaEx.Server)
+    send(worker_name, :stop_streaming)
   end
 
 #OTP API
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -48,7 +48,7 @@ defmodule KafkaEx.Server do
     {:reply, response, {metadata, client, event_pid}}
   end
 
-  def handle_call({:metadata, topic}, _from, {metadata, client, event_pid} = state) do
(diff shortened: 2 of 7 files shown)


lib/kafka_ex/server.ex
```diff
@@ -15,14 +15,22 @@ defmodule KafkaEx.Server do
     {:ok, {metadata, client, nil}}
   end
 
-  defp send_request(metadata, client, topic, partition, request_fn, parser, timeout \\ 100) do
+  defp send_request(metadata, client, topic, partition, request_fn, parser, timeout \\ 100, return_response \\ true) do
```
bjhaid: This function always has a return value,  `return_response` has added an extra arg plus extra lines of code, the only case where we cal `send_request` without the need for a response already _devnulls_ the response.

lib/kafka_ex/server.ex
```diff
@@ -80,10 +95,23 @@ defmodule KafkaEx.Server do
...
+    receive do
+      {:tcp, _, data} ->
+        {:ok, response} = KafkaEx.Protocol.Fetch.parse_response(data)
+        message_sets = response[topic][partition].message_set
+        if Enum.empty?(message_sets) do
+          :timer.sleep(1000)
+          start_stream(pid, metadata, client, handler, topic, partition, offset)
+        else
+          Enum.each(message_sets, fn(message_set) -> GenEvent.notify(pid, message_set) end)
+          last_offset = Enum.at(Enum.reverse(message_sets), 0).offset
```
bjhaid: I think we can abstract the logic here into a recursive function smartly returns the last_offset

e.g:

```elixir
defp notify_handler([], handler_pid, offset), do: offset

defp notify_handler([message_set|message_sets], handler_pid, offset) do
   GenEvent.notify(handler_pid, message_set)
   notify_handler(message_sets, handler_pid, message_set.offset)
end
```

that saves us the cost of doing `Enum.reverse` and we can conveniently remove the `if/else` block

lib/kafka_ex.ex
```diff
@@ -199,10 +199,15 @@ defmodule KafkaEx do
...
-    GenEvent.stream(pid)
+    stream      = GenServer.call(worker_name, {:create_stream, handler})
+    send(worker_name, {:start_streaming, topic, partition, offset, handler})
+    stream
+  end
+
+  @spec stop_streaming(Keyword.t) :: :ok
+  def stop_streaming(opts \\ []) do
+    worker_name = Keyword.get(opts, :worker_name, KafkaEx.Server)
+    send(worker_name, :stop_streaming)
```
bjhaid: :heart:

lib/kafka_ex.ex
```diff
@@ -199,10 +199,15 @@ defmodule KafkaEx do
     offset      = Keyword.get(opts, :offset, 0)
     handler     = Keyword.get(opts, :handler, KafkaExHandler)
 
-    {:ok, pid}  = GenEvent.start_link
-    :ok         = GenEvent.add_handler(pid, handler, [])
-    send(worker_name, {:start_streaming, topic, partition, offset, pid, handler})
-    GenEvent.stream(pid)
+    stream      = GenServer.call(worker_name, {:create_stream, handler})
+    send(worker_name, {:start_streaming, topic, partition, offset, handler})
```
bjhaid: :heart:

bjhaid: I am thinking to make `start_streaming` a first class function, that creates a worker that starts with an `event_pid` in its state so when the worker crashes it can be properly restarted with an `event_pid` though this would make sense when I have completed the `consumer_group` work that I am doing so the worker can commit offset and be smart to know where to continue streaming from after a crash.

lib/kafka_ex/server.ex
jacklund: :+1:

lib/kafka_ex/server.ex
jacklund: Yeah, it's probably less-than-great the way it is. I can refactor it.

lib/kafka_ex.ex
jacklund: :+1:

PR #22, opened 2015-05-06
Move modules around in place of funs and refactor send_request.

bjhaid:
I merged `send_request_and_return_response` and `send_request` into a single function, I am passing the module name around instead of a fun, that removes duplication in a couple of places.


jacklund:
:+1:

PR #24, opened 2015-06-18
Clearly defined datastructures for request and response

bjhaid:


jacklund:
A couple of minor things, but overall :+1:

Issue #27, opened 2015-06-29
Using KafkaEx term for config doesn't work when creating a release with exrm


joshrotenberg:
The application name in mix.exs is :kafka_ex but the calls to get_env (and the documentation) use KafkaEx as the key. This seems to work when run via iex -S mix in a project but breaks when I generate a release. See https://github.com/joshrotenberg/relthing for an example with errors. Changing the two calls to Applciation.get_env to use :kafka_env and using that in my config seems to solve the issue.

PR #28, opened 2015-06-29
rename config key from KafkaEx to :kafka_ex

joshrotenberg:
The config key should match the application name so that configs can be pulled correctly when running a built release (created via exrm).


joshrotenberg:
Fixes issue #27

bjhaid:
thanks :heart:

Issue #19, opened 2015-04-19
Add support for compression


jacklund:

dantswain:
@jacklund Has any work been done on this, or any design ideas discussed?  I have some ideas and might be able to subsidize some time since it would help me a lot at work :)

PR #35, opened 2015-09-18
Support reading snappy-compressed logs

dantswain:
This partially addresses #19 - I am only dealing with snappy and I split reading from writing to make a bite-sized change.  I plan to also work on writing.  Adding other compression formats shouldn't be hard once the protocol details are worked out. 

There's a few caveats here:

* I used a fork of the Erlang snappy lib; the only change is that the fork 'supports' Erlang R18 (it's just a check in rebar.config).  There is a PR open on the trunk repository and it should be trivial to switch back to the trunk repo once the PR is merged.

* Because of the way that compression works, if one fetches an offset that is inside a compressed batch of messages, the entire batch is returned.  This is the behavior for the Ruby client, Poseidon.  It wouldn't be _that_ hard to check the offsets of the batch and filter out ones that were not asked for.

* `KafkaEx.Util` could certainly use some refactoring.  I did not want to inflict my particular style and potentially introduce merge conflicts.  I'd be glad to have a go at it if you are OK with that.


bjhaid:
Thanks for the work, see my responses inline

There's a few caveats here:

I used a fork of the Erlang snappy lib; the only change is that the fork 'supports' Erlang R18 (it's just a check in rebar.config). There is a PR open on the trunk repository and it should be trivial to switch back to the trunk repo once the PR is merged.

  I am wary of NIFs but given that the available snappy implementations use NIFs then it's fine. 

Because of the way that compression works, if one fetches an offset that is inside a compressed batch of messages, the entire batch is returned. This is the behavior for the Ruby client, Poseidon. It wouldn't be that hard to check the offsets of the batch and filter out ones that were not asked for.

  I think the assumption you made here is fine


KafkaEx.Util could certainly use some refactoring. I did not want to inflict my particular style and potentially introduce merge conflicts. I'd be glad to have a go at it if you are OK with that.

  I agree with you here `KafkaEx.Util` is used only in `lib/kafka_ex/protocol/fetch.ex`, all the functions should be moved to `lib/kafka_ex/protocol/fetch.ex` and the module can be nuked.

Overall I think this is a great, next steps would be to include integrations test, and have compression for produce.

dantswain:
Thanks for the feedback, @bjhaid!

I'm looking at the producer API and it puts the compression property on the individual message, not the request:

```
@type t :: %Request{topic: binary, partition: integer, required_acks: binary, timeout: integer, messages: list}
# vs
@type t :: %Message{key: binary, value: binary, compression: atom}
```

I propose that the compression property be moved to the Request struct.  My reasoning is that Kafka suggests compressing message sets rather than individual messages in order to better take advantage of compression.

Do you want a new PR for sending compressed messages or should I add to this one?

bjhaid:
Compression should be on the message struct, see:

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets

A message can also recursively contain a message-set

Do you want a new PR for sending compressed messages or should I add to this one?

  I think they are related, as there is no easy way to have integration test for fetching compressed messages without producing compressed messages

Also a design decision we made that you have to keep in mind is that we assume a produce request would contain messages for one topic on the same partition (single partition).

Thanks for the effort!

bjhaid:
Thinking through it again, I agree with you on the compression being moved, my bad I wasn't thinking about the exposed API

dantswain:
Hah!  OK.  I was just in the process of writing out a big explanation of why I did it that way ;)  I was also referencing the same doc.

Writing is done and I've created an integration test that passes.  I'd like to take a pass at some refactoring.

lib/kafka_ex/protocol/produce.ex
```diff
@@ -22,32 +22,38 @@ defmodule KafkaEx.Protocol.Produce do
...
-    sub = << 0 :: 8, 0 :: 8 >> <> bytes(key) <> bytes(value)
+  defp create_message(value, key, attributes \\ 0) do
+    sub = << 0 :: 8, attributes :: 8-signed >> <> bytes(key) <> bytes(value)
     crc = :erlang.crc32(sub)
     << crc :: 32 >> <> sub
   end
 
-  defp bytes(nil), do: << 0 :: 32 >>
-
+  defp bytes(nil), do: << -1 :: 32-signed >>
```
dantswain: This was a bug.  `nil` is encoded as `255, 255, 255, 255` in the protocol.

dantswain:
@bjhaid As part of refactoring the Fetch module, I added a Message struct.  I'm currently calling `Map.from_struct` before returning because I did not want to break the API completely.  If you want, I can change it to return Message structs and update the documentation.

bjhaid:
I think adding the `Message` struct is fine, the difference the struct introduces is adding `__struct__` key to the map, which I don't think it's a behavior anyone should be depending on, so I don't think it breaks backwards compatibility :+1: for this

lib/kafka_ex/protocol/produce.ex
bjhaid: :heart:

dantswain:
Done!

dantswain:
A handful of the integration tests failed on Travis; I got similar failures on one laptop but not on a different one.  I get different test failures on different runs, so seems like something finicky wrt async?

bjhaid:
The failures are usually related to consumer_groups, it probably needs some delay(but I have not being able to identify how long this delay is) for kafka to identify the coordinator/have all nodes in sync, that's why the test suite is run twice in travis.

Thanks for this, I would merge this soon

bjhaid:
@dantswain thanks published a new version of kafka_ex with your compression feature https://hex.pm/packages/kafka_ex/0.2.0

PR #39, opened 2015-09-23
Fix handling chunked snappy-compressed data

### LINKED ISSUE #38: Snappy not handling chunked compression data
When the message size gets over around 1kb, snappy uses chunked compression and the decompress operation is not working correctly.

I should have a patch for this within the next day or so (it's blocking me at work).

dantswain:
Fixes #38


dantswain:
Thanks!

PR #41, opened 2015-09-23
Change Produce.Message defaults to nil

### LINKED ISSUE #40: KafkaEx.Protocol.Produce.Message default values should be nil
I'm not sure if there's really a difference in the messages that are produced, but it does cause problems with testing when the parsing correctly parses a nil value.

dantswain:
Closes #40 

Tests still pass


dantswain:
Thanks!

PR #42, opened 2015-09-26
Remove snappy from applications, update README

### LINKED ISSUE #37: Snappy not resolving as a dependency
With a new project and

```
  def application do
    [applications: [:logger, :kafka_ex]]
  end

  defp deps do
    [{:kafka_ex, "~> 0.2"}]
  end
end
```

in the `mix.exs` file, I do a `mix deps.get`, `mix deps.compile`, and `iex` fails with

```
** (Mix) Could not start application snappy: could not find application file: snappy.app
```

Inspection of `mix.lock` shows that snappy was not resolved as a dependency.

I'm not sure why this is happening - I put snappy in the KafkaEx `mix.exs`.  Manually adding snappy as a dependency to the downstream project should resolve it in an emergency, but this should be fixed :(

dantswain:
Should close #37.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,15 @@ And run:
 mix deps.get
 ```
 
+*Note* If you wish to use snappy for compression or decompression, you
+ must add
+ [snappy-erlang-nif](https://github.com/fdmanana/snappy-erlang-nif) to
+ your project's mix.exs.  Note the main repository does not support
+ Erlang R18; there is a branch at
+ https://github.com/ricecake/snappy-erlang-nif
+ that supports R18.  That branch is included in KafkaEx's mix.exs for
+ development and testing.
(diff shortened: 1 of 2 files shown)


dantswain:
@bjhaid you may want to push a hotfix to hex after this.  0.2.0 will fail in the wild unless the user adds snappy to their mix.exs :(

bjhaid:
@dantswain thanks for this, but it would be great if you can include an example of how to add it to the `mix.exs` file

dantswain:
Sorry @bjhaid I got bogged down for the last two weeks with ElixirConf and work travel.  Thanks for updating!

bjhaid:
that's fine, thanks for the work :heart:

PR #44, opened 2015-10-11
Use Application.get_env instead of Mix.Config.read

dantswain:
This returns the same value and is The OTP Way to do it.

Also removed the `TestHelper.get_bootstrap_hosts` function that no longer appears to be used.


bjhaid:
Thanks! :heart: would merge as soon as I can

PR #46, opened 2015-10-11
Fix consumer group offset commit

### LINKED ISSUE #45: AutoCommit of offsets does not work for custom consumer groups
I think the issue lies here (note the `consumer_group` field of the request is not set): https://github.com/kafkaex/kafka_ex/blob/bbba99187d30be01e77057c075b7a379fbcbe386/lib/kafka_ex/server.ex#L278

I'm making changes to the integration test to verify and then try to fix.

dantswain:
This should fix #45 

I changed this one specific test not to use a randomly generated topic name.  I think this is better.  Given time I'd update all of the tests.

I still get one test failing on consumer groups but it was failing before.

```
  1) test fetch does not commit offset with auto_commit is set to false (KafkaEx.ConsumerGroup.Test)
     test/integration/consumer_group_test.exs:102
     ** (ArgumentError) argument error
     stacktrace:
       :erlang.hd([])
       test/integration/consumer_group_test.exs:105
```

### DIFF
--- a/lib/kafka_ex/protocol/offset.ex
+++ b/lib/kafka_ex/protocol/offset.ex
@@ -7,6 +7,21 @@ defmodule KafkaEx.Protocol.Offset do
   defmodule Response do
     defstruct topic: "", partition_offsets: []
     @type t :: %Response{topic: binary, partition_offsets: list}
+
+    @doc """
+    Returns the numeric offset of the first partition in the response,
+    or nil if the topic is not found.
+
+    This is a convenience for the most typical case.
+    """
+    @spec first_partition_offset(atom | t) :: number | nil
+    def first_partition_offset(:topic_not_found) do
+      nil
+    end
+    def first_partition_offset([%Response{partition_offsets: partition_offsets}]) do
+      first_partition = hd(partition_offsets)
+      first_partition.offset |> hd
+    end
   end
 
   def create_request(correlation_id, client_id, topic, partition, time) do
--- a/lib/kafka_ex/protocol/offset.ex
+++ b/lib/kafka_ex/protocol/offset.ex
@@ -7,21 +7,6 @@ defmodule KafkaEx.Protocol.Offset do
   defmodule Response do
     defstruct topic: "", partition_offsets: []
     @type t :: %Response{topic: binary, partition_offsets: list}
-
-    @doc """
-    Returns the numeric offset of the first partition in the response,
-    or nil if the topic is not found.
-
-    This is a convenience for the most typical case.
-    """
-    @spec first_partition_offset(atom | t) :: number | nil
-    def first_partition_offset(:topic_not_found) do
-      nil
-    end
-    def first_partition_offset([%Response{partition_offsets: partition_offsets}]) do
-      first_partition = hd(partition_offsets)
-      first_partition.offset |> hd
-    end
   end
 
   def create_request(correlation_id, client_id, topic, partition, time) do
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -275,7 +275,12 @@ defmodule KafkaEx.Server do
             last_offset = response |> hd |> Map.get(:partitions) |> hd |> Map.get(:last_offset)
             case last_offset do
               nil -> {response, state}
-              _ -> {_, state} = offset_commit(state, %Proto.OffsetCommit.Request{topic: topic, offset: last_offset})
+              _ ->
+                offset_commit_request = %Proto.OffsetCommit.Request{
+                  topic: topic,
+                  offset: last_offset,
+                  consumer_group: consumer_group(state)}
+                {_, state} = offset_commit(state, offset_commit_request)
                 {response, state}
             end
           _    -> {response, state}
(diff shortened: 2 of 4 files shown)


lib/kafka_ex/server.ex
```diff
@@ -284,6 +289,7 @@ defmodule KafkaEx.Server do
   end
 
   defp offset_commit(state, offset_commit_request) do
+    IO.puts(inspect offset_commit_request)
```
bjhaid: :) I am sure you forgot this here

lib/kafka_ex/server.ex
```diff
@@ -292,8 +298,16 @@ defmodule KafkaEx.Server do
 
     offset_commit_request_payload = Proto.OffsetCommit.create_request(state.correlation_id, @client_id, offset_commit_request)
     response = KafkaEx.NetworkClient.send_sync_request(broker, offset_commit_request_payload) |> Proto.OffsetCommit.parse_response
+    IO.puts(inspect response)
```
bjhaid: and this one too

lib/kafka_ex/server.ex
```diff
@@ -292,8 +298,16 @@ defmodule KafkaEx.Server do
...
 
     {response, %{state | correlation_id: state.correlation_id+1}}
   end
 
+  defp consumer_group(state) do
+    if state.consumer_group == false do
+      "kafka_ex"
```
bjhaid: Use the module attribute `@consumer_group` here instead

lib/kafka_ex/protocol/offset.ex
```diff
@@ -7,6 +7,21 @@ defmodule KafkaEx.Protocol.Offset do
...
+    This is a convenience for the most typical case.
+    """
+    @spec first_partition_offset(atom | t) :: number | nil
+    def first_partition_offset(:topic_not_found) do
+      nil
+    end
+    def first_partition_offset([%Response{partition_offsets: partition_offsets}]) do
+      first_partition = hd(partition_offsets)
+      first_partition.offset |> hd
+    end
```
bjhaid: if this function is used only in `TestHelper` I would rather make it a private function in `TestHelper`

bjhaid:
I prefer we keep using randomly generated topic names, so multiple runs of test don't get _influenced_ by the internal state of kafka

lib/kafka_ex/protocol/offset.ex
dantswain: I think this is debatably useful externally.  I pulled this in from test code I had already written in another application where I'm using KafkaEx.

This gets at a bigger question, though - why do all of these structs support arrays when the KafkaEx API doesn't really support making a request for multiple offsets of multiple topics/partitions?

dantswain:
What kind of influence are you concerned about?  Shouldn't the tests expose those?

Conversely:
* I was able to get a lot of useful debugging information by running `kafka-console-consumer.sh` against the topic that the test uses.
* I'm concerned that some of the tests implicitly rely on the fact that the topic is empty.
* It's not easy to delete topics in kafka, so the test broker gets polluted with random topics.
* Fixed topic names and dealing with populated topics is the "normal" use case, where we're reading from a specific topic that's being populated by some external service.  I feel that the test suite should mirror the normal use case.

lib/kafka_ex/protocol/offset.ex
bjhaid: > I think this is debatably useful externally. I pulled this in from test code I had already written in another application where I'm using KafkaEx.

You should not rely on any function residing in the protocols, they _can_ change (I also don't intend exposing any function in the protocol)

> This gets at a bigger question, though - why do all of these structs support arrays when the KafkaEx API doesn't really support making a request for multiple offsets of multiple topics/partitions?

The plan was to support multiple topics/partition, we never just got there (or wasn't sure the library users would find it useful), the protocols mostly support it, supporting it has the complexity of topic/partition combination not belonging to the same leader. I have thought of just piping the response API functions to `hd` and returning the struct, just haven't been able to make up my mind on that.

bjhaid:
You can use

> I was able to get a lot of useful debugging information by running kafka-console-consumer.sh against the topic that the test uses.

You can always print the topic name to console and do same

>I'm concerned that some of the tests implicitly rely on the fact that the topic is empty.

Yes, we should be able to guarantee that the library would behave correctly against a pristine kafka cluster, for use cases we don't have test against I would appreciate a PR

>It's not easy to delete topics in kafka, so the test broker gets polluted with random topics.

Use [install_kafka.sh](https://github.com/kafkaex/kafka_ex/blob/master/install_kafka.sh) to spin up a test kafka instance and [cleanup.sh](https://github.com/kafkaex/kafka_ex/blob/master/cleanup.sh) to clean up after you.

> Fixed topic names and dealing with populated topics is the "normal" use case, where we're reading from a specific topic that's being populated by some external service. I feel that the test suite should mirror the normal use case.

Agreed there is a limit to how much you can mirror this use case given we want to be able to run the test with limited external dependencies.


We have ran into a couple of bugs in the past, that were a result of us reusing topic names, except the test requires it, I wouldn't want us to reuse topic names.

Thanks as usual!

lib/kafka_ex/protocol/offset.ex
dantswain: I kind of agree with you on not exposing any functions from inside the protocol. But I think the same argument kind of applies to exposing data structures from inside the protocol.  This helper function was intended to at least hide the details of the underlying data structure.

I can move this function, it's not a huge deal.  But I think if you want to (rightly) encapsulate the details of the protocol, we should find a way to expose the data via an API.  Maybe publicly-facing data structures to encapsulate the salient details that shouldn't change.

dantswain: This function also mirrors `KafkaEx.Protocol.OffsetFetch.Response.last_offset/1`, though that function name masks the possibility of multiple partitions.

ibizaman:
My 2 cents about having an empty topic or not.
Apart when explicitly needing an offset of 0, we could first fetch the latest_offset, then store, then fetch/stream from the old latest_offset + 1. That being said, I don't if we would gain something doing that.
Either way, this is already done in some tests.

PR #49, opened 2015-11-23
create_stream should handle multiple calls

beedub:
The test case shows what was failing for me. I'm newish to elixir and kafka, so the code and/or kafka usage may be wrong

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -127,11 +127,14 @@ defmodule KafkaEx.Server do
   end
 
   def handle_call({:create_stream, handler}, _from, state) do
-    state = unless state.event_pid && Process.alive?(state.event_pid) do
+    if state.event_pid && Process.alive?(state.event_pid) do
+      info = Process.info(self)
+      Logger.log(:warn, "'#{info[:registered_name]}' already streaming handler '#{handler}'")
+    else
       {:ok, event_pid}  = GenEvent.start_link
-      %{state | event_pid: event_pid}
+      state = %{state | event_pid: event_pid}
+      :ok = GenEvent.add_handler(state.event_pid, handler, [])
     end
-    :ok = GenEvent.add_handler(state.event_pid, handler, [])
     {:reply, GenEvent.stream(state.event_pid), state}
   end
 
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/server.ex
```diff
@@ -127,11 +127,13 @@ defmodule KafkaEx.Server do
   end
 
   def handle_call({:create_stream, handler}, _from, state) do
-    state = unless state.event_pid && Process.alive?(state.event_pid) do
+    if state.event_pid && Process.alive?(state.event_pid) do
+      Logger.log(:warn, "Stream handler already exists")
```
bjhaid: You should include the name of the handler in the message logged.

bjhaid:
Thanks for this, once you update the logging I will merge it.

lib/kafka_ex/server.ex
```diff
@@ -127,11 +127,14 @@ defmodule KafkaEx.Server do
   end
 
   def handle_call({:create_stream, handler}, _from, state) do
-    state = unless state.event_pid && Process.alive?(state.event_pid) do
+    if state.event_pid && Process.alive?(state.event_pid) do
+      info = Process.info(self)
+      Logger.log(:warn, "#{info[:registered_name]} already streaming")
```
beedub: @bjhaid i updated the log msg to include the name of the process, not the handler, as it seems to be more related than the handler involved. does that make sense?

bjhaid: I would still include the handler in the log msg because the error can come from the default process which can be confusing

lib/kafka_ex/server.ex
```diff
@@ -127,11 +127,14 @@ defmodule KafkaEx.Server do
   end
 
   def handle_call({:create_stream, handler}, _from, state) do
-    state = unless state.event_pid && Process.alive?(state.event_pid) do
+    if state.event_pid && Process.alive?(state.event_pid) do
+      info = Process.info(self)
+      Logger.log(:warn, "'#{info[:registered_name]}' already streaming handler '#{handler}'")
```
beedub: @bjhaid how's this?

bjhaid: :+1: thanks! would merge later today

PR #53, opened 2015-11-29
Add optional initial argument for stream handler

### LINKED ISSUE #52: Allow init arguments for stream handler
It would be awesome if we could give custom arguments to GenEvent handlers started by calling [`stream`](https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex.ex#L244).

This way I could add an initial state for example, or a callback function to be called by a custom handler.

Would you accept a PR doing that ?

ibizaman:
This fixes #52.

As a side note: I followed all steps in the contribution document and there were dialyzer errors but not on the lines I modified.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -223,6 +223,7 @@ defmodule KafkaEx do
   - worker_name: the worker we want to run this metadata request through, when none is provided the default worker `KafkaEx.Server` is used
   - offset: When supplied the fetch would start from this offset, otherwise would start from the last committed offset of the consumer_group the worker belongs to. For Kafka < 0.8.2 you should explicitly specify this.
   - handler: the handler we want to handle the streaming events, when none is provided the default KafkaExHandler is used
+  - handler_arg: argument to give to the handler - in other words its initial state
   - auto_commit: specifies if the last offset should be commited or not. Default is true. For Kafka < 0.8.2 set this to false.
 
 
@@ -245,6 +246,7 @@ defmodule KafkaEx do
     worker_name = Keyword.get(opts, :worker_name, KafkaEx.Server)
     offset      = Keyword.get(opts, :offset)
     handler     = Keyword.get(opts, :handler, KafkaExHandler)
+    handler_arg = Keyword.get(opts, :handler_arg, [])
     auto_commit = Keyword.get(opts, :auto_commit, true)
 
     offset = case offset do
@@ -257,7 +259,7 @@ defmodule KafkaEx do
       _   -> offset
     end
 
-    stream      = GenServer.call(worker_name, {:create_stream, handler})
+    stream      = GenServer.call(worker_name, {:create_stream, handler, handler_arg})
     send(worker_name, {:start_streaming, topic, partition, offset, handler, auto_commit})
     stream
   end
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -223,7 +223,7 @@ defmodule KafkaEx do
   - worker_name: the worker we want to run this metadata request through, when none is provided the default worker `KafkaEx.Server` is used
   - offset: When supplied the fetch would start from this offset, otherwise would start from the last committed offset of the consumer_group the worker belongs to. For Kafka < 0.8.2 you should explicitly specify this.
   - handler: the handler we want to handle the streaming events, when none is provided the default KafkaExHandler is used
-  - handler_arg: argument to give to the handler - in other words its initial state
+  - handler_init: initial state for the handler - leave the default value [] when using the default handler
   - auto_commit: specifies if the last offset should be commited or not. Default is true. For Kafka < 0.8.2 set this to false.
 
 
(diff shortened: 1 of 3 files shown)


lib/kafka_ex.ex
```diff
@@ -245,6 +246,7 @@ defmodule KafkaEx do
     worker_name = Keyword.get(opts, :worker_name, KafkaEx.Server)
     offset      = Keyword.get(opts, :offset)
     handler     = Keyword.get(opts, :handler, KafkaExHandler)
+    handler_arg = Keyword.get(opts, :handler_arg, [])
```
bjhaid: I will call this `handler_args`

lib/kafka_ex.ex
```diff
@@ -223,6 +223,7 @@ defmodule KafkaEx do
   - worker_name: the worker we want to run this metadata request through, when none is provided the default worker `KafkaEx.Server` is used
   - offset: When supplied the fetch would start from this offset, otherwise would start from the last committed offset of the consumer_group the worker belongs to. For Kafka < 0.8.2 you should explicitly specify this.
   - handler: the handler we want to handle the streaming events, when none is provided the default KafkaExHandler is used
+  - handler_arg: argument to give to the handler - in other words its initial state
```
bjhaid: should be

> list of arguments

and not

> argument

lib/kafka_ex/server.ex
```diff
@@ -126,14 +126,14 @@ defmodule KafkaEx.Server do
     {:reply, metadata, state}
   end
 
-  def handle_call({:create_stream, handler}, _from, state) do
+  def handle_call({:create_stream, handler, handler_arg}, _from, state) do
```
bjhaid: I will add a guard to confirm `handler_args` is a list

bjhaid:
Thanks for this, but I think the approach of having an initial argument will make this less of a generic solution (see [KafkaExHandler](https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/handler.ex)). A more generic approach will be to pass the `handler_args` along with the message in a `keyword list` or a `map` and let the handler destructure the list/map to get the message and args and determine what to do with them, that IMO is more generic.

ibizaman:
> A more generic approach will be to pass the handler_args along with the message in a keyword list or a map and let the handler destructure the list/map to get the message and args and determine what to do with them
I must say I don't see what you mean, sorry. I've got the impression this is what the handler_args does, no? 
Do you mean this ?

```elixir
defmodule KafkaExHandler do
  use GenEvent

  def handle_event({message_set, handler_args}, message_sets) do
    {:ok, [message_set|message_sets]}
  end

  # ...
end
```

bjhaid:
> Do you mean this ?

Something like that, this is what I meant:

```elixir
defmodule KafkaExHandler do
  use GenEvent

  def handle_event(%{message: message_set, handler_args: handler_args}, message_sets) do
    {:ok, [message_set|message_sets]}
  end

  # ...
end
```

ibizaman:
Okay, but I don't see how this is more generic.
For example, what if I don't want to keep the ever growing list of messages in the message_sets? Well, you could ignore the handle_event's second argument, but then why imposing it?

With using anything as the initial state, you can do more IMO. Notably changing the state like hereunder. I don't see how you could do that by passing `handler_args`.

```elixir
defmodule MyHandler do
  use GenEvent

  def handle_event(message, my_state) do
    do_something(message)
    new_state = change_state(my_state)
    {:ok, new_state}
  end
end
```

bjhaid:
> Well, you could ignore the handle_event's second argument

True and return whatever you want as your state.

>  but then why imposing it?

It is not imposed, with your approach the `handle_event` function in an handler that does something different from what you intend doing (e.g `KafkaHandler.handle_event/2`) will have to keep track of state somehow to detect when the second argument is the initial state and when it isn't.

ibizaman:
How would you go implementing this with your method?

```elixir
defmodule MyHandler do
  use GenEvent

  def handle_event(message, my_state) do
    do_something(message)
    new_state = change_state(my_state)
    {:ok, new_state}
  end
end
```

And about detecting the initial call from others, we could use the [`init/1`](http://elixir-lang.org/docs/stable/elixir/GenEvent.html#c:init/1) callback, no?

bjhaid:
> How would you go implementing this with your method?

```elixir
defmodule MyHandler do
  use GenEvent

  def handle_event(%{message: message, handler_args: handler_args}, _) do
    do_something(message)
    do_another_thing(handler_args)
    {:ok, []}
  end
end
```

> And about detecting the initial call from others, we could use the init/1 callback, no?

This is fine by me if you change the variable to name to `handler_init` and make the test reflect that the `handler_init` is used in the `init` function and not in the `handle_event` function as the test currently reflects. Also this use case is slightly different from the one I have been proposing.

ibizaman:
From a syntax point of view, I don't really care if this is implemented "your" way or "my" way. The only thing worrying me is the functionality. I got the impression `handler_args` is a constant in your case: you cannot use the result of `do_another_thing` for the next call, correct?

> This is fine by me if you change the variable to name to handler_init and make the test reflect that the handler_init is used in the init function and not in the handle_event function as the test currently reflects. Also this use case is slightly different from the one I have been proposing.

I completely agree, the name are badly chosen and I think this thread reflects that quite well. :)
Does the following commit suits you?

bjhaid:
Thanks!, I will merge as soon as I can. :heart:

Issue #54, opened 2015-11-30
Fetch offset


beedub:
[KafkaEx.fetch](https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex.ex#L148) fetches messages using the last_offset provided by the last_offset api. Kafka looks to be returning messages starting from that offset (as opposed to starting after that offset), meaning that fetches will always include the last message consumed. I would have expected that the consumer would only fetch new messages.

1. Is this the expected behavior? Do other kafka libraries do this?
1. Should my code be tracking the offsets instead?

ibizaman:
Hi, I don't know about 1. but about 2. you could simply use `last_offset + 1` instead of tracking the offset yourself.

bjhaid:
@beedub sorry for my late response

> Is this the expected behavior? Do other kafka libraries do this?

No we should be moving the offset by 1 (`last_offset + 1`)

> Should my code be tracking the offsets instead?

No, please send in a PR to fix this if you can or use @ibizaman suggestion as I stop gap till I can fix it.

Thanks!

beedub:
@bjhaid working on the changes now

PR #55, opened 2015-12-03
fetch should use offset + 1

beedub:
fixes https://github.com/kafkaex/kafka_ex/issues/54

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -149,7 +149,7 @@ defmodule KafkaEx do
         if last_offset <= 0 do
           0
         else
-          last_offset
+          last_offset + 1
         end
       _   -> offset
     end
@@ -255,7 +255,7 @@ defmodule KafkaEx do
         if last_offset <= 0 do
           0
         else
-          last_offset
+          last_offset + 1
         end
       _   -> offset
     end
--- a/test/integration/consumer_group_test.exs
+++ b/test/integration/consumer_group_test.exs
@@ -100,10 +100,12 @@ defmodule KafkaEx.ConsumerGroup.Test do
   end
 
   test "fetch does not commit offset with auto_commit is set to false" do
-    KafkaEx.create_worker(:fetch_no_auto_commit_worker)
-    Enum.each(1..10, fn _ -> KafkaEx.produce(%Proto.Produce.Request{topic: "food", required_acks: 1, messages: [%Proto.Produce.Message{value: "hey"}]}) end)
-    offset = KafkaEx.fetch("food", 0, offset: 0, worker: :fetch_no_auto_commit_worker, auto_commit: false) |> hd |> Map.get(:partitions) |> hd |> Map.get(:message_set) |> Enum.reverse |> hd |> Map.get(:offset)
-    offset_fetch_response = KafkaEx.offset_fetch(:fetch_no_auto_commit_worker, %Proto.OffsetFetch.Request{topic: "food"}) |> hd
+    topic = generate_random_string
+    worker_name = :fetch_no_auto_commit_worker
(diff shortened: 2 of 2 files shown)


lib/kafka_ex.ex
```diff
@@ -144,14 +144,10 @@ defmodule KafkaEx do
...
-        if last_offset <= 0 do
-          0
-        else
-          last_offset
-        end
-      _   -> offset
+    unless offset do
+      # retrieve last stored offset from kafka
+      last_offset = offset_fetch(worker_name, %KafkaEx.Protocol.OffsetFetch.Request{topic: topic, partition: partition}) |> KafkaEx.Protocol.OffsetFetch.Response.last_offset
+      offset = max(last_offset + 1, 0)
```
beedub: simplified the logic here. is the floor for the value from `offset_fetch` -1? if so, then this line can be simplified to `offset = last_offset`

bjhaid: the logic here

`offset = max(last_offset + 1, 0)` is wrong, if `last_offset` is `0` `offset` is set to `1` and thus you will miss the first message, also I prefer the `case` over `if/unless` (`if/unless` expands/compiles to case statement)

test/integration/integration_test.exs
```diff
@@ -192,6 +192,28 @@ defmodule KafkaEx.Integration.Test do
...
+
+    # Produce a new message
+    produce_response =  KafkaEx.produce(%Proto.Produce.Request{topic: random_string, required_acks: 1, messages: [%Proto.Produce.Message{value: "bye foo"}]}) |> hd
+    # There's a new message, so fetch should return it
+    fetch_response = KafkaEx.fetch(random_string, 0) |> hd
+    message = fetch_response.partitions |> hd |> Map.get(:message_set) |> hd
+    assert message.value == "bye foo"
+    assert message.offset == 1
+  end
+
```
bjhaid: this test should be in [`KafkaEx.ConsumerGroup.Test`](https://github.com/kafkaex/kafka_ex/blob/master/test/integration/consumer_group_test.exs) as offset_management is `consumer_group` related and not available in kafka < 0.8.2.

Also there are test cases that covers all this behavior (wrongly) that you should modify in [`KafkaEx.ConsumerGroup.Test`](https://github.com/kafkaex/kafka_ex/blob/master/test/integration/consumer_group_test.exs)

lib/kafka_ex.ex
beedub: if `last_offset` returns 0, then that means that another process consumed message 0 and committed that offset, so 1 in that case would be right? i'm new to kafka, so apologies if i'm missing something

>  I prefer the case over if/unless

For this specific piece of code, I actually find it difficult to read because it requires a scan of 7 additional lines of code to see that `offset`'s value isn't changed if it's not nil; the `unless` makes this immediately apparent. I don't want to make this a point of contention though and will use `case` if that's what you want

test/integration/integration_test.exs
beedub: there's actually a test that covers this in `ConsumerGroup.Test` that I'll just updated instead

lib/kafka_ex.ex
bjhaid: `last_offset` can return 0 if it a new `consumer_group` and none of the consumers in the group has consumed anything.

bjhaid:
@beedub Bump, I intend publishing a release today and intend this to go into the release, do you need help?

beedub:
will take a look now. when i last left off, i was pretty close, but will let you know in the next hour

beedub:
@bjhaid updated. this also includes changes from https://github.com/kafkaex/kafka_ex/pull/58

PR #57, opened 2015-12-06
Fixes error code reported by update_consumer_metadata

ronanh:
update_consumer_metadata always reported error code 0...


bjhaid:
This PR is wrong, `update_consumer_metadata` does not always report error_code 0, it's a recursive function that will try 3 times and report failure [here](https://github.com/kafkaex/kafka_ex/blob/ce9fd81e5e1b7df39b422a1852fdc13b3d42ed62/lib/kafka_ex/server.ex#L192)

ronanh:
Please re-read the code, the recursion starts with `error_code` = 0 and remains unchanged until the end...

what you want to recurse with is `response.error_code`

bjhaid:
You are correct, my bad. Thanks for catching this, will merge shortly

PR #59, opened 2015-12-16
document overriding options when creating worker

### LINKED ISSUE #50: Why use worker names instead of PIDs ?
Hi,

First of all thank you for this great library!

I was wondering why do you require a name for kafka_workers? Why not being simply anonymous? `KafkaEx.create_worker` returns a PID after all.
This can be cumbersome for implementing a pool of workers for example. And furthermore it is considered bad practice to create a lot of atoms since they live forever in memory.

ibizaman:
closes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,10 +73,20 @@ iex> Application.put_env(:kafka_ex, :brokers, [uris: [{"localhost", 9092}, {"loc
 :ok
 ```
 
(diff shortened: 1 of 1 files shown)


bjhaid:
I think we should have a list of options that `create_stream` takes and what they do, though this is a good start. thanks!

Issue #60, opened 2015-12-16
Make elixir handle kafka daemons


ibizaman:
For now testing the app means first launching 3 separate kafka daemons by hand and also cleaning up by hand. Furthermore, there is no way to tweak the ports used by the daemons. In my case, one of the port clashes with the "production" server.

I propose to make elixir setup and monitor the kafka daemons. I don't see any cons for that apart from the coding time. The pros are:
* dynamic configuration of ports in the test setup;
* automatic setup of one or multiple kafka daemons;
* having control over the lifetime of the daemon;
* being able to clean the state of the kafka server when we want to;
* and not the least - making it easy for users of the library to setup their own tests with a kafka daemon.

I already worked on that for my project - not finished yet - so I'm asking mainly if you are interested by a PR with this?

bjhaid:
You can send in a PR and lets see how it looks.

ibizaman:
Hi,

I have a branch working with this right now but I still get errors in the tests. Those errors are I suppose related to your comment [here](https://github.com/kafkaex/kafka_ex/issues/56#issuecomment-162022778) because by running the tests twice (and of course stopping the kafka daemons in between), different test case fail.

With this PR, the kafka daemon are started before the tests and stopped after the tests, so we cannot anymore run the tests twice.

Would you have more information on what could solve this issue? I will test by adding an arbitrary delay between the "create_worker" and "fetch" as you suggest.

bjhaid:
I might have to see your branch to have an idea of what you are doing, but I would assume making them mix tasks should be adequate, also kafka 0.9.0 introduces some new API that should solve some of the `consumer_groups` woes

ibizaman:
This exists, although I do not know if that can be of any help though: https://github.com/chbatey/kafka-unit.

dantswain:
IMHO this is a lot of code to handle a task that isn't within the scope of KafkaEx.  The maintenance burden is a concern - it could, for example, lead to KafkaEx breaking on changes to the server.properties file that have nothing to do with the operation of the client itself.  Perhaps it could spin off to a new project?

ibizaman:
Well, I switched to docker now and it offers a very nice alternative. I mean, the burden's on docker now. I use the wurstmeister/kafka image with docker-compose and it works perfectly!

dantswain:
Cool.  Maybe this issue can be closed?

bjhaid:
I am closing this.

PR #61, opened 2016-01-06
Make elixir handle kafka daemons

ibizaman:
Instead of using the install_script to launch the kafka daemons with nearly hardcoded values, now they are launched from the tests themselves and can be configured for the tests.
It should close issue #60 although it is not ready yet.


bjhaid:
I am slightly worried that this is a lot of non-unit-tested code to spin up test dependencies. While kafka is a required dependency I do not intend to tie spinning of up of kafka to the test code, it takes away freedom from whoever wants to spin up kafka in a different manner to run tests.

If we are to replace the scripts, I will prefer mix task(s) that just shell out do about the same thing the current  set of scripts do.

ibizaman:
Okay, that seems nice too.

In tests I want to write for my application, I need to delete a topic and create it anew. How would you do that having mix tasks?

bjhaid:
Sorry took me a while to respond, the kafka API does not expose any means to deleting topics, but [google search](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=delete%20kafka%20topic) shows there might be tricks to deleting kafka topics

bjhaid:
Closing this based on discussion on https://github.com/kafkaex/kafka_ex/issues/60

PR #64, opened 2016-02-04
Allow creating unnamed server workers

localshred:
Creating named workers is useful but falls down if I want to create a
large group of workers to work with (for instance) a poolboy pool.

This is a very naive attempt at changing as little code as possible to
introduce an atom `:no_name` which will cause the
`KafkaEx.Server.start_link` method to not register any name for the
started process.

I wrote a simple description in the README as well. I'm happy to update
the description or change the implementation per your guidance.

Thanks for the library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,24 @@ iex> KafkaEx.create_worker(:pr, [uris: uris, consumer_group: "kafka_ex", consume
 {:ok, #PID<0.172.0>}
 ```
 
+### Create KafkaEx worker
+
+You may find you want to create many workers, say in conjunction with
+a `poolboy` pool. In this scenario you usually won't want to name these worker processes.
+
+To create an unnamed worked with `create_worker`:
+```elixir
+iex> KafkaEx.create_worker(:no_name) # indicates to the server process not to name the process
+{:ok, #PID<0.171.0>}
+```
+
+To create an unnamed worker directly with the `KafkaEx.Supervisor`:
+```elixir
+iex> worker_init = [] # whatever options you want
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,6 +83,24 @@ iex> KafkaEx.create_worker(:pr, [uris: uris, consumer_group: "kafka_ex", consume
...
+a `poolboy` pool. In this scenario you usually won't want to name these worker processes.
+
+To create an unnamed worked with `create_worker`:
+```elixir
+iex> KafkaEx.create_worker(:no_name) # indicates to the server process not to name the process
+{:ok, #PID<0.171.0>}
+```
+
+To create an unnamed worker directly with the `KafkaEx.Supervisor`:
+```elixir
```
bjhaid: I am not sure we want to document creating workers with `KafkaEx.Supervisor` even though it's what we do under the covers, it can be confusing

bjhaid:
Thanks! I will merge asap

README.md
localshred: Ok. I can remove that. Just wanted to provide docs for the rationale, but didn't seem appropriate to include a poolboy example. Shall I remove?

bjhaid: That's fine, if you have a working example with poolboy I will appreciate if you can PR that separately in an examples folder.

Thanks!

localshred: Ok so I'll leave as is and change it in another PR.

bjhaid:
merged! let me know if you want a release with this patch.

localshred:
I'm testing from my fork still so I'll let you know if I missed anything. Thanks!!!

PR #66, opened 2016-02-07
Config timeout

barruumrex:
Add a sync_timeout configuration. Defaults to 1000ms which is what it was before. Can be set in the environment config or directly in the worker spec. Worker spec takes precedence over environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,8 @@ In your config/config.exs add the list of kafka brokers as below:
 ```elixir
 config :kafka_ex,
   brokers: [{HOST, PORT}],
-  consumer_group: consumer_group #if no consumer_group is specified "kafka_ex" would be used as the default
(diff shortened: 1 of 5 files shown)


bjhaid:
Thanks, I generally avoid generating docs till I create a release, as you can see there are things not related to your PR in the docs, so I'll suggest you revert that commit

bjhaid:
I'll merge once you revert as suggested

barruumrex:
Cool. Glad I generated those as a separate commit. Thanks for all the help. :)

bjhaid:
Released https://hex.pm/packages/kafka_ex/0.4.0

barruumrex:
oh. Awesome! I was expecting to have to run off my branch for at least a couple days. Thank you very much. :100:

Issue #68, opened 2016-02-15
Support optional startup of default worker


dantswain:
In my production scenario, I need to be able to run multiple consumer groups (potentially across multiple broker clusters) from the same VM.  

I can probably accomplish this by breaking my consumer groups up and starting one via the default `:kafka_ex` worker on startup and the rest as part of my boot process.  This seems klunky, though.  Ideally I think one could optionally start the application with no worker servers running and then launch their own set of workers on demand and name them as desired.

I think this is also contributing to some of the problems I'm seeing with consumer group management.  I think the existence of the `:kafka_ex` worker by default has made it easier to miss bugs in `KafkaEx.Server` - if we set up a test case and forget to route everything through the test worker, we could find that the default worker is handling some request that we did not expect it to handle and therefore masking what should really be a test failure.

bjhaid:
Submit a PR that makes turning off the default worker(`:kafka_ex`) configurable

dantswain:
OK, I will.  Just wanted to make sure you were amenable to the idea.

I'm getting failing integration tests on master.  The first time I ran it there were 5, then 1, then 4... :(  Is this known?

bjhaid:
This https://github.com/kafkaex/kafka_ex/pull/35#issuecomment-141843295 is still true, I guess they should mostly be sorted with the new Consumer Group API, but it's not completely going away since we have to support `ConsumerGroupMetadata` forever

dantswain:
I'm seeing something different.  I get tests passing after the first try if I only run one broker, but with 3 they don't pass after one run.

I think this is because I have my test cluster set up to replicate topics (replicating the environment of our production cluster).  Probably there are async issues with writing to one broker and being told to read from another.  Without replication, required acks kind of solve the async problem because you're always reading from the same broker to which you wrote and the ack guarantees the message is already available.  With replication, that assumption fails and occasionally you're being told to read from a replica that is not yet in sync.  I'm not 100% sure on that, but it's my best guess.

So the bottom line is probably that the integration tests are good enough right now, but in the future you/we may want to modify them to work against an async environment.  It's a pain in the butt, but it can be done.

Or, at minimum, the documentation should be updated to state that the integration tests assume no replication.

bjhaid:
This is closed by #78

bjhaid:
@dantswain I just published a new release([0.5.0](https://github.com/kafkaex/kafka_ex/releases/tag/0.5.0)), thanks for all the work and patches

PR #69, opened 2016-02-17
Remove default values from structs

dantswain:
This is a lead up to #68 but does not implement that feature.  I think that this will potentially solve a number of bugs, as well.  The default value of "kafka_ex" for the consumer group in many places was causing unset values to be missed.  Generally speaking, I don't think structs should have default values unless those default values mean something specific (error codes, specific unset values, empty arrays, etc.).  Empty strings as defaults, for example, can mask a bug where a value is not being set properly.

Unfortunately, I didn't write a failing test before I started this, but I know that it has solved at least one bug I was experiencing when using a custom worker with a custom consumer group.

Summary of changes:
- Fix root repo in CONTRIBUTING.md
- Remove 'false' consumer group (it's not well defined what this even means)
- Shift offset fetch into Server.ex - this reduces some repeated code and should be more performant
- Add a worker_consumer_group helper function
- Remove default values from many structs, fix a couple bugs that that uncovered
- Remove KafkaEx.Protocol.Fetch.Request struct, which was not used by any other code
- Give the Server.ex module a State substruct, which makes pattern matching easier
- Add guards to several functions - these could theoretically be removed, but they were very helpful in working out bugs.

Sorry for the big PR; all of the tests still pass and I have not introduced any new dialyzer warnings.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -5,4 +5,4 @@ use Mix.Config
       {"localhost", 9093},
       {"localhost", 9094},
     ],
-    consumer_group: false
+    consumer_group: "kafka_ex_dev"
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -31,10 +31,10 @@ defmodule KafkaEx do
   def create_worker(name, worker_init \\ [])
 
   def create_worker(name, worker_init) do
-    worker_init = case worker_init do
-      [] -> [uris: Application.get_env(:kafka_ex, :brokers)]
-      _   -> worker_init
-    end
+    defaults = [uris: Application.get_env(:kafka_ex, :brokers),
+                consumer_group: Application.get_env(:kafka_ex, :consumer_group)]
+
+    worker_init = Keyword.merge(defaults, worker_init)
 
     Supervisor.start_child(KafkaEx.Supervisor, [worker_init, name])
   end
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -149,16 +149,6 @@ defmodule KafkaEx do
     max_bytes     = Keyword.get(opts, :max_bytes, @max_bytes)
     auto_commit   = Keyword.get(opts, :auto_commit, true)
 
-    offset = case offset do
-      nil -> last_offset = offset_fetch(worker_name, %KafkaEx.Protocol.OffsetFetch.Request{topic: topic, partition: partition}) |> KafkaEx.Protocol.OffsetFetch.Response.last_offset
-        if last_offset <= 0 do
-          0
-        else
-          last_offset + 1
-        end
-      _   -> offset
-    end
-
     GenServer.call(worker_name, {:fetch, topic, partition, offset, wait_time, min_bytes, max_bytes, auto_commit})
   end
 
@@ -172,9 +162,7 @@ defmodule KafkaEx do
 
   @spec offset_fetch(atom, KafkaEx.Protocol.OffsetFetch.Request.t) :: KafkaEx.Protocol.OffsetFetch.Response.t
   def offset_fetch(worker_name,
-                   offset_fetch_request = %Protocol.OffsetFetch.Request{
-                     consumer_group: consumer_group})
-  when is_binary(consumer_group) and byte_size(consumer_group) > 0 do
+                   offset_fetch_request = %Protocol.OffsetFetch.Request{}) do
     GenServer.call(worker_name, {:offset_fetch, offset_fetch_request})
   end
 
@@ -261,27 +249,7 @@ defmodule KafkaEx do
     handler_init = Keyword.get(opts, :handler_init, [])
     auto_commit  = Keyword.get(opts, :auto_commit, true)
 
-    get_consumer_group = fn() -> worker_consumer_group(worker_name) end
-    consumer_group = Keyword.get_lazy(opts, :consumer_group, get_consumer_group)
-
-    offset = case offset do
-               nil ->
-                 request = %Protocol.OffsetFetch.Request{topic: topic,
-                                                         partition: partition,
-                                                         consumer_group: consumer_group}
-
-                 last_offset = offset_fetch(worker_name, request)
-                 |> KafkaEx.Protocol.OffsetFetch.Response.last_offset
-
-                 if last_offset <= 0 do
-                   0
-                 else
-                   last_offset + 1
-                 end
-               _   -> offset
-             end
-
-    stream      = GenServer.call(worker_name, {:create_stream, handler, handler_init})
+    stream = GenServer.call(worker_name, {:create_stream, handler, handler_init})
     send(worker_name, {:start_streaming, topic, partition, offset, handler, auto_commit})
     stream
   end
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -3,6 +3,8 @@ defmodule KafkaEx do
   @type uri() :: [{binary|char_list, number}]
   @type worker_init :: [{:uris, uri}, {:consumer_group, binary|false}, {:sync_timeout, non_neg_integer}]
 
+  alias KafkaEx.Protocol
+
   @doc """
   create_worker creates KafkaEx workers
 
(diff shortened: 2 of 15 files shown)


dantswain:
I also think that there is an off-by-one bug floating around in the offset commit code.  The [last commit](https://github.com/dantswain/kafka_ex/commit/edb5a14fe79144cba6f7a2d378c0d9dadf8057b5) makes no difference in integration test results.  I can try to tackle that in a followup PR.

dantswain:
I just realized i wasn't including the consumer group tests... Will fix those ASAP.

dantswain:
OK.  All of the tests pass now.

dantswain:
:rage1: They pass when I run them locally.

dantswain:
Not sure what the test failures in travis were about - looks like timeouts, could it be lag on the travis side? They went away after the most recent commit and I've run them multiple times locally with the provided test cluster (install_kafka.sh).

config/config.exs
```diff
@@ -5,4 +5,4 @@ use Mix.Config
       {"localhost", 9093},
       {"localhost", 9094},
     ],
-    consumer_group: false
+    consumer_group: "kafka_ex_dev"
```
bjhaid: this will break for users using kafka < 0.8.2, see https://github.com/kafkaex/kafka_ex/commit/214cac1881883dd3ba85c075fbc76bda7d66dafa

bjhaid:
There are too many changes in this PR, I would prefer you split this into smaller chunks and PR them separately.

lib/kafka_ex.ex
```diff
@@ -29,10 +31,10 @@ defmodule KafkaEx do
...
 
   def create_worker(name, worker_init) do
-    worker_init = case worker_init do
-      [] -> [uris: Application.get_env(:kafka_ex, :brokers)]
-      _   -> worker_init
-    end
+    defaults = [uris: Application.get_env(:kafka_ex, :brokers),
+                consumer_group: Application.get_env(:kafka_ex, :consumer_group)]
+
+    worker_init = Keyword.merge(defaults, worker_init)
```
bjhaid: This is a stylistic change and I prefer the original version as it is easy to know what's happening without looking at `Keyword` docs, also the original version will outperform the change

bjhaid: taking a second pass through this, I see that the intent is different from the original implementation, but based on explanation elsewhere this will break support for kafka < 0.8.2 also, there is an implicit assumption that you can pass a completely different set of kafka brokers when creating a worker that this will also break

dantswain: I expect that if I call `create_worker(:name, [keyword list])`, it will merge the keyword list with defaults (the uris and consumer group set in the config).  The original version does not work that way at all.  With the original version, if I call `create_worker(:name, consumer_group: "foo")` or `create_worker(:name, [consumer_group: "foo"])` (they are equivalent), the worker will fail to start because the uris will not be populated.  I would have to call `create_worker(:name, consumer_group: "foo", uris: [....])`, which seems odd if I want to use the uris that are already present in my App config.

I thought `Keyword.merge/2` was fair game here, since the function accepts a `Keyword.t` as its argument.  If the function name isn't clear enough, I could add a comment.

I don't understand your comment about performance at all.  What are your performance concerns with this?

dantswain:
I can try to break this down into smaller PRs now that I've worked through the kinks.  I would really appreciate quick turn around on those, though, as this fixes some fairly serious bugs.  Consumer groups with names other than "kafka_ex" do not work at all in the master branch.

config/config.exs
dantswain: Is there a way to test that support?  From what I can tell, the consumer group setting here was being inconsistently honored throughout the code.  In some places, the default value of "kafka_ex" was being used.

lib/kafka_ex.ex
dantswain: This does not break the ability to pass a separate set of brokers.

bjhaid:
> Consumer groups with names other than "kafka_ex" do not work at all in the master branch.

This spec https://github.com/kafkaex/kafka_ex/blob/master/test/integration/consumer_group_test.exs#L44-L87 uses a different consumer_group name and it's fine as far as I can tell

> I would really appreciate quick turn around on those.

I will try my best to respond as soon as I can, but remember I am human :)

config/config.exs
bjhaid: Those code paths _should not_ (consumer groups did not fully exist prior to kafka 0.8.2) be called by users of kafka < 0.8.2, I assumed this was documented but I probably missed it.

dantswain:
I know, I wrote that test ;)   IIRC (it's been a couple days), the behavior that I was seeing is that the offset would be committed properly but the offset fetch was not using the consumer group.  So if I would run `kafka-consumer-offset-checker.sh` I would see the offset incrementing, but then when I call `KafkaEx.fetch`, I still get the log starting from the beginning.  That led me to find the consumer group not being set correctly in the offset_fetch request, which led me to realize that the default struct values were causing this to be missed.

lib/kafka_ex.ex
bjhaid: I understand what `Keyword.merge/2` does, first pass through the code made it look like a stylistic change (since there are a couple of them in this PR), but we still have to support users not using `consumer_groups` hence the code without the default `consumer_group` is not different from the new implementation

config/config.exs
dantswain: I see what you mean now.  If the consumer group is falsey then the consumer metadata update timer is not started.

This is sort of minor, but IMO `nil` makes more sense as a default.  That way the type is `binary | nil`.

lib/kafka_ex.ex
bjhaid: Can you please write a test for this, I just found a bug we did not discuss, passing a `consumer_group` without passing `uris` breaks the original implementation

bjhaid:
I just created a worker from console and I can confirm that `KafkaEx.fetch` with the worker I created correctly increments the offset see: https://gist.github.com/bjhaid/7382f3b0759b0269f08e

lib/kafka_ex.ex
dantswain: That was why I changed it in the first place ;P

>  With the original version, if I call create_worker(:name, consumer_group: "foo") or create_worker(:name, [consumer_group: "foo"]) (they are equivalent), the worker will fail to start because the uris will not be populated. I would have to call create_worker(:name, consumer_group: "foo", uris: [....]), which seems odd if I want to use the uris that are already present in my App config.

bjhaid: Sorry it's past 8pm CT, brains don't give optimal performance :p

dantswain: 9:30p ET and I've been staring at this all day for two days straight :scream: ;)

I'm making a PR for this change alone.  It has some interesting complications.

dantswain:
So I'm not sure why, the example you pasted is setting the consumer group to "kafka_ex" (at least it is when I run it, copy and pasted).  https://gist.github.com/dantswain/d517a57eac7cd19f2c07

Confirmed by the kafka-consumer-group-offset-checker.sh:
```
$ kafka-consumer-offset-checker.sh --zookeeper localhost:2181 --topic foo --group bar
Exiting due to: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /consumers/bar/offsets/foo/0.

dswain at itsacomputer in ~
$ kafka-consumer-offset-checker.sh --zookeeper localhost:2181 --topic foo --group kafka_ex
Group           Topic                          Pid Offset          logSize         Lag             Owner
kafka_ex        foo                            0   5               1000            995             none
```

dantswain:
Heh.  It's actually _overwriting_ the consumer group:

```
iex(1)> KafkaEx.create_worker(:foo, consumer_group: "bar", uris: Application.get_env(:kafka_ex, :brokers))
{:ok, #PID<0.138.0>}
iex(2)> :sys.get_state(:foo).consumer_group
"bar"
iex(3)> KafkaEx.fetch("foo", 0, worker_name: :foo, max_bytes: 100)
[%KafkaEx.Protocol.Fetch.Response{partitions: [%{error_code: 0,
     hw_mark_offset: 1000, last_offset: 8,
     message_set: [%KafkaEx.Protocol.Fetch.Message{attributes: 0,
       crc: 1505099747, key: "", offset: 6, value: "bar7"},
      %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3372933746, key: "",
       offset: 7, value: "bar8"},
      %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3188576996, key: "",
       offset: 8, value: "bar9"}], partition: 0}], topic: "foo"}]
iex(4)> :sys.get_state(:foo).consumer_group
"kafka_ex"
```

bjhaid:
https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/server.ex#L114 is the bug, this is terrible bug, I will patch ASAP

dantswain:
Good find. I don't think this is the only bug. I'm relatively certain that this PR fixes another one. I'm sure you want to get on this bug asap, but if you can review my other PRs tonight I can spend time tomorrow getting at the others.

bjhaid:
Is there anything yet to be merged from this PR?

dantswain:
Yes - the "kafka_ex" default values in some structs (as discussed in #74).  I also moved the offset fetch into server.ex, but that might not be totally necessary (though may be more performant).

bjhaid:
Can you make separate PRs for those. Thanks!

dantswain:
Will do.  Tomorrow :)   I was waiting for the ones from today to clear.

dantswain:
I'm going to close this.  All of the bugs I know of will have been fixed as of #77.  There were a couple ancillary changes in this PR but they are not critical at all AFAIK.

PR #71, opened 2016-02-18
Fix parameter merging

dantswain:
Another breakout from #69 

Fixes default parameter merging as we discussed.  This means that the `@consumer_group` module attribute within server.ex should no longer be relevant.  However, I found that in some of the integration tests, using `false` as the consumer group by default would cause a failure.  Upon further inspection, it appears that this is due to `auto_commit` defaulting to true in `KafkaEx.fetch`.

So, should we also set `auto_commit` to default to false?  Or include logic within server.ex to not allow offset commits unless the consumer group name is set?  I'm not familiar with Kafka < 0.8.2, but AFAIU it doesn't make sense to attempt to commit an offset unless one is using consumer groups (the docs don't appear to show support for that https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetCommitRequest).

### DIFF
--- a/test/integration/integration_test.exs
+++ b/test/integration/integration_test.exs
@@ -49,12 +49,20 @@ defmodule KafkaEx.Integration.Test do
     assert consumer_group_update_interval == 30000
   end
 
-  test "create_worker provides a default consumer_group of 'kafka_ex'" do
+  test "create_worker provides a default consumer_group of false" do
     {:ok, pid} = KafkaEx.create_worker(:baz, uris: uris)
     consumer_group = :sys.get_state(pid).consumer_group
 
-    assert consumer_group == "kafka_ex"
+    assert consumer_group == false
   end
+
+  test "create_worker allows us to provide a consumer group" do
+    {:ok, pid} = KafkaEx.create_worker(:bah, consumer_group: "my_consumer_group")
+    consumer_group = :sys.get_state(pid).consumer_group
+
+    assert consumer_group == "my_consumer_group"
+  end
+
 
   test "create_worker takes a consumer_group option and sets that as the consumer_group of the worker" do
     {:ok, pid} = KafkaEx.create_worker(:joe, [uris: uris, consumer_group: "foo"])
--- a/test/integration/integration_test.exs
+++ b/test/integration/integration_test.exs
@@ -49,11 +49,11 @@ defmodule KafkaEx.Integration.Test do
     assert consumer_group_update_interval == 30000
   end
 
-  test "create_worker provides a default consumer_group of false" do
+  test "create_worker provides a default consumer_group of 'kafka_ex'" do
     {:ok, pid} = KafkaEx.create_worker(:baz, uris: uris)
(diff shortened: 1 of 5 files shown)


test/integration/integration_test.exs
```diff
@@ -49,13 +49,21 @@ defmodule KafkaEx.Integration.Test do
     assert consumer_group_update_interval == 30000
   end
 
-  test "create_worker provides a default consumer_group of 'kafka_ex'" do
```
bjhaid: Providing a default `consumer_group` of `kafka_ex` allows us to magically manage for users of the library, which makes kafka_ex work out of the box with little intervention. I prefer to make disabling `consumer_group` a second class thing so you have to explicitly not want it

dantswain: I think that's fine, but I think it should be a separate setting - see the complication that I point out above.  Maybe something like

```
config :kafka_ex,
  consumer_group: "kafka_ex",
  disable_consumer_group: false
```

bjhaid: I think the example you showed is why disabling `consumer_group` is based off of `consumer_group` configuration also requiring this to be explicitly configured means we require users to configure this to get the magical offset management. This might be better as explicit is better than implicit, but then we have to document this new behavior we are introducing, also we should change https://github.com/dantswain/kafka_ex/blob/fix_parameter_merging/lib/kafka_ex.ex#L38-L39 to use `Application.fetch_env!` instead of `Application.get_env` since this is a require param

dantswain: Actually, I could argue that magically doing anything for the user doesn't help them understand Kafka, which is already fairly murky.  But I think it's within your prerogative as a maintainer to want to continue to provide a default.  

I do, however, think the kinks need to be worked out.  If you want the default to be "kafka_ex", then that should be the setting in the config.exs file.  Disabling consumer groups by setting the name of the consumer group to false is indirect and I think that's evidenced by the complication we see in server.ex, where the consumer group's name is used and it's entirely possible to have autocommit set to true in conditions when consumer groups should theoretically be unsupported.

dantswain: OK, let me know what you want at a high level and I'll take a stab at implementing it.  How do you think the config.exs should be laid out so that the name of the consumer group and whether or not consumer groups are enabled is clear to the user?

bjhaid: I suppose we should require `consumer_group` by default and it's either a binary (string) or an atom `:no_consumer_group` (you can chose a more fitting name here), so if it's not passed in `worker_init` or included in `config.exs` then we should blow up on creating the worker.

dantswain:
@bjhaid Updated as per our conversation.  Note I added some assertions to ensure that a consumer group is set when trying to use functions that would require it.  The change from `false` to `:no_consumer_group` could break some folks' existing config.exs files.

bjhaid:
Thanks I will look at this after CoB today

PR #73, opened 2016-02-18
Introduce Server.State module

dantswain:
This is admittedly a stylistic change, but it's more consistent with style in the rest of the code.  I use this pattern heavily myself and find it to be very useful.

Note this means removing the defaults from the State struct (the module attributes don't cross module boundaries), but the defaults are already getting set inside `Server.init/1` and they should really only be set in one place.


dantswain:
Another re: #69 
Could have minor merge conflicts with #71 - I will take responsibility for that.

bjhaid:
No worries I'll merge it

PR #74, opened 2016-02-18
Remove defaults for topic and partition

dantswain:
This was masking at least one bug - the partition was not being passed through to the offset commit request in Server.fetch, which would probably cause offset commits to always be commit to partition 0.

This looks like a lot of changes, but most of the changes are either removing the default values from structs or updating the tests where they were previously relying on the defaults.  I also tried to update documentation wherever it was missing values for partition or topic.

I would like to do a similar PR for the default "kafka_ex" consumer group, which I believe is masking similar bugs as per #69.  I'll wait until #71 is resolved for that.

### DIFF
--- a/lib/kafka_ex/protocol/offset_commit.ex
+++ b/lib/kafka_ex/protocol/offset_commit.ex
@@ -1,11 +1,11 @@
 defmodule KafkaEx.Protocol.OffsetCommit do
   defmodule Request do
-    defstruct consumer_group: "kafka_ex", topic: "", partition: 0, offset: 0, metadata: ""
+    defstruct consumer_group: "kafka_ex", topic: nil, partition: nil, offset: nil, metadata: ""
     @type t :: %Request{consumer_group: binary, topic: binary, partition: integer, offset: integer}
   end
 
   defmodule Response do
-    defstruct partitions: [], topic: ""
+    defstruct partitions: [], topic: nil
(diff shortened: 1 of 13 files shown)


lib/kafka_ex/protocol/offset_commit.ex
```diff
@@ -1,11 +1,11 @@
 defmodule KafkaEx.Protocol.OffsetCommit do
   defmodule Request do
-    defstruct consumer_group: "kafka_ex", topic: "", partition: 0, offset: 0, metadata: ""
+    defstruct consumer_group: "kafka_ex", topic: nil, partition: nil, offset: nil, metadata: ""
```
bjhaid: `kafka_ex` should go away too

lib/kafka_ex/protocol/offset_fetch.ex
```diff
@@ -1,11 +1,11 @@
 defmodule KafkaEx.Protocol.OffsetFetch do
   defmodule Request do
-    defstruct consumer_group: "kafka_ex", topic: "", partition: 0
+    defstruct consumer_group: "kafka_ex", topic: nil, partition: nil
```
bjhaid: same here too

lib/kafka_ex/protocol/offset_commit.ex
dantswain: I agree.  I wanted to do that in a separate PR because it has some other implications.

PR #76, opened 2016-02-19
KafkaEx workers overrides the consumer_group to "kafka_ex" everytime

bjhaid:
KafkaEx.offset_fetch and KafkaEx.offset_commit is called, this ensures
that the consumer_group is consistent and not overriden.

@dantswain please review. Thanks

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -48,6 +48,7 @@ defmodule KafkaEx.Server do
     {correlation_id, metadata} = metadata(brokers, 0, sync_timeout)
     state = %State{metadata: metadata, brokers: brokers, correlation_id: correlation_id, consumer_group: consumer_group, metadata_update_interval: metadata_update_interval, consumer_group_update_interval: consumer_group_update_interval, worker_name: name, sync_timeout: sync_timeout}
     {:ok, _} = :timer.send_interval(state.metadata_update_interval, :update_metadata)
+    Logger.log(:debug, "consumer_group: #{consumer_group}, worker_name: #{name}")
 
     if consumer_group do
       {:ok, _} = :timer.send_interval(state.consumer_group_update_interval, :update_consumer_metadata)
@@ -109,9 +110,13 @@ defmodule KafkaEx.Server do
     {:reply, response, state}
   end
 
+  def handle_call({:offset_fetch, offset_fetch}, _from, state = %State{consumer_group: false}) do
+    {:reply, :topic_not_found, state}
+  end
+
   def handle_call({:offset_fetch, offset_fetch}, _from, state) do
     {broker, state} = case Proto.ConsumerMetadata.Response.broker_for_consumer_group(state.brokers, state.consumer_metadata) do
-      nil -> {_, state} = update_consumer_metadata(state, offset_fetch.consumer_group)
+      nil -> {_, state} = update_consumer_metadata(state)
         {Proto.ConsumerMetadata.Response.broker_for_consumer_group(state.brokers, state.consumer_metadata), state}
       broker -> {broker, state}
     end
@@ -138,7 +143,7 @@ defmodule KafkaEx.Server do
   end
 
   def handle_call({:consumer_group_metadata, consumer_group}, _from, state) do
-    {consumer_metadata, state} = update_consumer_metadata(state, consumer_group)
+    {consumer_metadata, state} = update_consumer_metadata(state)
     {:reply, consumer_metadata, state}
   end
 
@@ -194,7 +199,7 @@ defmodule KafkaEx.Server do
   end
 
   def handle_info(:update_consumer_metadata, state) do
-    {:noreply, update_consumer_metadata(state, state.consumer_group) |> elem(1)}
+    {:noreply, update_consumer_metadata(state) |> elem(1)}
   end
 
   def handle_info(_, state) do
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/server.ex
```diff
@@ -48,6 +48,7 @@ defmodule KafkaEx.Server do
     {correlation_id, metadata} = metadata(brokers, 0, sync_timeout)
     state = %State{metadata: metadata, brokers: brokers, correlation_id: correlation_id, consumer_group: consumer_group, metadata_update_interval: metadata_update_interval, consumer_group_update_interval: consumer_group_update_interval, worker_name: name, sync_timeout: sync_timeout}
     {:ok, _} = :timer.send_interval(state.metadata_update_interval, :update_metadata)
+    Logger.log(:debug, "consumer_group: #{consumer_group}, worker_name: #{name}")
```
dantswain: Stray debug log

lib/kafka_ex/server.ex
```diff
@@ -109,9 +110,13 @@ defmodule KafkaEx.Server do
     {:reply, response, state}
   end
 
+  def handle_call({:offset_fetch, offset_fetch}, _from, state = %State{consumer_group: false}) do
```
dantswain: This should get handled by #71 - it will actually error out, which I think is better.  You don't want someone to find out too late that they're operating under invalid assumptions.

bjhaid: will update this to point to the correct `consumer_group` `:no_consumer_group`

lib/kafka_ex/server.ex
bjhaid: thanks will remove

lib/kafka_ex/server.ex
```diff
@@ -209,21 +214,21 @@ defmodule KafkaEx.Server do
     Enum.each(state.brokers, fn(broker) -> KafkaEx.NetworkClient.close_socket(broker.socket) end)
   end
 
-  defp update_consumer_metadata(state, consumer_group), do: update_consumer_metadata(state, consumer_group, 3, 0)
+  defp update_consumer_metadata(state), do: update_consumer_metadata(state, 3, 0)
```
dantswain: What's the significance of the 3 here?  Looks like a number of tries?  Should it be a module attribute?

lib/kafka_ex/server.ex
```diff
@@ -313,9 +318,13 @@ defmodule KafkaEx.Server do
     end
   end
 
+  defp offset_commit(state = %State{consumer_group: false}, offset_commit_request) do
```
dantswain: This will also get handled by #71 and I also think it should error out.

lib/kafka_ex/server.ex
bjhaid: yes should be a module attribute and yes it's the number of retries

lib/kafka_ex/server.ex
```diff
@@ -209,21 +214,21 @@ defmodule KafkaEx.Server do
...
 
-  defp update_consumer_metadata(state, consumer_group, retry, _error_code) do
-    consumer_group_metadata_request = Proto.ConsumerMetadata.create_request(state.correlation_id, @client_id, consumer_group)
+  defp update_consumer_metadata(state = %State{consumer_group: consumer_group, correlation_id: correlation_id}, retry, _error_code) do
+    consumer_group_metadata_request = Proto.ConsumerMetadata.create_request(correlation_id, @client_id, consumer_group)
     data = Enum.find_value(state.brokers, fn(broker) -> KafkaEx.NetworkClient.send_sync_request(broker, consumer_group_metadata_request, state.sync_timeout) end)
     response = Proto.ConsumerMetadata.parse_response(data)
     case response.error_code do
-      0 -> {response, %{state | consumer_metadata: response, consumer_group: consumer_group, correlation_id: state.correlation_id + 1}}
+      0 -> {response, %{state | consumer_metadata: response, correlation_id: state.correlation_id + 1}}
```
dantswain: It makes more sense to me to have this function just return the mutated state, and everywhere you have `{consumer_metadata, state} = update_consumer_metadata(state)` you could just have `state = update_consumer_metadata(state)` and then use `state.consumer_metadata`.

test/integration/consumer_group_test.exs
```diff
@@ -7,10 +7,11 @@ defmodule KafkaEx.ConsumerGroup.Test do
 
   test "consumer_group_metadata works" do
     random_string = generate_random_string
-    pid = Process.whereis(KafkaEx.Server)
     KafkaEx.produce(%Proto.Produce.Request{topic: "food", required_acks: 1, messages: [%Proto.Produce.Message{value: "hey"}]})
     KafkaEx.fetch("food", 0, offset: 0)
-    metadata = KafkaEx.consumer_group_metadata(KafkaEx.Server, random_string)
+    KafkaEx.create_worker(:consumer_group_metadata_worker, consumer_group: random_string, uris: Application.get_env(:kafka_ex, :brokers))
+    pid = Process.whereis(:consumer_group_metadata_worker)
+    metadata = KafkaEx.consumer_group_metadata(:consumer_group_metadata_worker, random_string)
```
dantswain: I realize this isn't actually something that you changed in this  PR, but there's at least three different "metadata"s involved in a Kafka client - consumer metadata, consumer group metadata, and offset metadata.  Variables named just "metadata" therefore carry some risk of confusion...

lib/kafka_ex/server.ex
bjhaid: all the `consumer_group: false` will be updated to `consumer_group: :no_consumer_group`, also this is a hack, I am beginning to think `KafkaEx.offset_fetch` and `KafkaEx.offset_commit` should not be an exposed API

dantswain:
Functionally looks fine except the forgotten debug log.  Style comments are to be taken with a grain of salt and are just my opinions ;)

test/integration/consumer_group_test.exs
bjhaid: The names were chosen to keep them close to the API they expose from Kafka

dantswain: I just meant the local variable "metadata".  Not a big deal.

PR #78, opened 2016-02-19
Allow disabling the default worker, update README and document config.exs

dantswain:
The motivation to disable the default worker is described in #68

I updated the README to point to config/config.exs for the application
env, and updated config/config.exs with comments describing how to use
each of the supported env settings.  I think this makes it easier to
keep everything up-to-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,19 +54,12 @@ end
 
 ### Configuration
 
-In your config/config.exs add the list of kafka brokers as below:
-```elixir
-config :kafka_ex,
-  brokers: [{HOST, PORT}],
-  consumer_group: consumer_group, #if no consumer_group is specified "kafka_ex" would be used as the default
-  sync_timeout: 1000 #Timeout used synchronous requests from kafka. Defaults to 1000ms.
-```
-
-Alternatively from iex:
-```elixir
-iex> Application.put_env(:kafka_ex, :brokers, [uris: [{"localhost", 9092}, {"localhost", 9093}], consumer_group: "kafka_ex"])
-:ok
-```
+See [config/config.exs](config/config.exs) for a description of
+configuration variables, including the Kafka broker list and default
+consumer group.  See
+http://elixir-lang.org/getting-started/mix-otp/distributed-tasks-and-configuration.html#application-environment-and-configuration
+for general info if you are unfamiliar with OTP application
+environments.
 
 You can also override options when creating a worker, see below.
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,8 +1,20 @@
 use Mix.Config
-  config :kafka_ex,
-    brokers: [
-      {"localhost", 9092},
-      {"localhost", 9093},
-      {"localhost", 9094},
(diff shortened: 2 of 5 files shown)


bjhaid:
I suggest this be 2 different PRs one for the config update and the other to enable removing the defauilt_worker, since they are sort of separate concerns, but it looks great, though I will suggest you have a spec that is not run in parallel to update application config with `disable_default_worker: true` and one verifying the default behaviour, otherwise I am :+1:

dantswain:
I figured they were small enough changes it wouldn't be a big deal - I had started to update the README for `disable_default_worker` and kind of said "screw it".

I'll take a stab at writing a test.  It should be possible to stop the app, set the env var, start it, make sure the supervisor has no child processes, then put everything back the way it was.

bjhaid:
I don't think you can stop the app, you can mutate the application config with [`Application.put_env`](http://elixir-lang.org/docs/v1.0/elixir/Application.html#put_env/4) but this will leave it modified, so the test cannot run in parallel and you have to ensure the test cleans up after itself

dantswain:
You _can_ stop the app ;)  See the added test.

dantswain:
Maybe Travis is going to prove me wrong.  I've definitely done this before with other projects.  I'm trying to reproduce the failures locally.

dantswain:
@bjhaid Did you trigger a rebuild?  I couldn't reproduce the failures locally using the same random seed.

bjhaid:
I triggered a rebuild

bjhaid:
So I have been able to make it fail again on travis, I will look at this more closely later today

bjhaid:
I think this spec should run completely independently, so you should give it a `moduletag` and include the tag in this [proplist](https://github.com/kafkaex/kafka_ex/blob/master/test/test_helper.exs#L3), then we can run the spec as `mix test --include integration --include consumer_group; mix test --include integration --include consumer_group && mix test test/kafka_ex_test.exs` and you should update [travis](https://github.com/kafkaex/kafka_ex/blob/master/.travis.yml#L8)

dantswain:
@bjhaid I will do that if you require it, but I really don't think it would help.   I don't think ExUnit sandboxes the tagged modules or anything - running `mix test --include integration --include consumer_group` doesn't do anything different than removing the tags and just running `mix test` AFAIK.   I've used application restarts a bunch of times in the past to test this kind of thing.  Can you tell me what your concerns are?  

I _did_ introduce some test failures when I added the sync_timeout value to config.exs (it was assuming `Application.get_env(:kafka_ex, :sync_timeout)` would return nil).  Those are fixed in the most recent commit.

bjhaid:
my suggestion is to run it as an isolated OS process and not rely on exunit to sandbox it (see the `&&`). My concerns was around Application restart and another spec running in parallel while the application is being restarted, I am not sure if `async: false` is per spec-file or takes effect globally. If `async: false` is global then it should be fine, I will try merging/rebasing against master and running it locally.

PR #81, opened 2016-02-24
Typespec fixes

dantswain:
Several of the typespecs were incorrect.  This fixes several of them to the best of my knowledge.  I ended up adding some specs in a couple places to help me figure out what the specs for KafkaEx should be.

Note that in Elixir v1.1+ there's a dialyzer warning in metadata.ex on the return value of `Port.info/1`.  As of v1.1 `Port.info/1` translates `:undefined` into `nil`.  I left the catch for `:undefined` so as not to break compatibility with Elixir 1.0.

The only remaining dialyzer warnings are the usual ones that get thrown for struct access.

### DIFF
--- a/lib/kafka_ex/compression.ex
+++ b/lib/kafka_ex/compression.ex
@@ -44,8 +44,6 @@ defmodule KafkaEx.Compression do
     {:ok, compressed_data} = :snappy.compress(data)
     {compressed_data, @snappy_attribute}
   end
-
-  @spec compress(compression_type_t, binary) :: {binary, attribute_t}
   def compress(:gzip, data) do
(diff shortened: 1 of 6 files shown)


lib/kafka_ex/compression.ex
```diff
@@ -44,8 +44,6 @@ defmodule KafkaEx.Compression do
     {:ok, compressed_data} = :snappy.compress(data)
     {compressed_data, @snappy_attribute}
   end
-
-  @spec compress(compression_type_t, binary) :: {binary, attribute_t}
```
dantswain: This was redundant with the spec in the above function clause.

bjhaid:
this PR has been merged

PR #85, opened 2016-03-08
Add consumer_group reader for introspection

dantswain:
In an application with potentially many workers, this can help with
consistency as other code may need to know the consumer group.  One
alternative is to pass around the name of the consumer group wherever we
pass around the worker name/pid, but this is more error prone than being
able to introspect the given process to determine its consumer group.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -46,6 +46,16 @@ defmodule KafkaEx do
   end
 
   @doc """
+  Returns the consumer group name for the given worker.
+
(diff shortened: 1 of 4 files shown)


dantswain:
:rage1: Travis passed on my fork.  I don't think the failure was related.  https://travis-ci.org/dantswain/kafka_ex/builds/114544090

I wonder if there's a way to get more info from the segfault.

bjhaid:
You should kick it again, I assume the segfault is related to docker which travis uses and not the code itself, I have noticed newer versions of docker segfault frequently. I will also look at the PR later today

Issue #65, opened 2016-02-06
Multi-Partition Timeout


barruumrex:
We're attempting to setup a group of workers to watch all of our partitions for certain events, transform them, and dump them into rethinkdb. We're running into an issue where certain workers are timing out and not being correctly restarted by the supervisor. Are we running into a bug with the supervision strategy or are we just doing this wrong? Would appreciate your assistance when you've got the time. 

The code looks like [this](https://gist.github.com/barruumrex/d4204f44b99c1912ad60)
The failure errors look like [this](https://gist.github.com/barruumrex/5e0cbb6329fac61f1c1c)
The observer tree looks like [this](http://i.imgur.com/iQmddln.png)

barruumrex:
Realized shortly after writing this that the workers are getting restarted correctly, it's that the stream doesn't get restarted. I assume that the correct thing to do is to setup my own application and link my processes to the kafka_ex workers. Link each process so that it dies when its associated worker dies, it also dies. When they get restarted, they re-establish the stream.

bjhaid:
I was writing a response when you responded, :+1:  to your approach.

barruumrex:
So we finished changing over to supervised workers that start and process the steam. Unfortunately the error that was killing the workers earlier happens a lot. We had to bump up the restart count in order to prevent kafka_ex from just bringing down the application. I plan to spend some time tomorrow digging in to figure out where that error is coming from. Hopefully we'll have a pull request soon to go along with the issue.

bjhaid:
From the stacktrace you shared the error is happening somewhere here https://github.com/kafkaex/kafka_ex/blob/0.3.0/lib/kafka_ex/network_client.ex#L28-L46, network related I guess

barruumrex:
Yeah. Looks like it's just timeout errors. Probably just a problem with us being on the vpm. It's solid as a rock now that I've recompiled that file with a longer default timeout. Is there a way to set the default timeout based on configuration environment? Would be nice to set the higher for dev/test and then drop it back low for prod.

bjhaid:
You can PR making the worker spec configurable

barruumrex:
PR is up. Let me know if there's anything else I need to do with it.

bjhaid:
The PR is great, what you have modified is better than what I was proposing ([worker_spec](https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/supervisor.ex#L13)). See my comments on the PR

PR #92, opened 2016-03-15
Improve error messages in network_client

dantswain:
- More consistently refer to brokers
- Specify async case in error message
- Specify send/receive in sync case

@bjhaid Quick review/merge would be much appreciated as this is related to a problem we're experiencing in production.  We're getting timeouts from the network client but can't tell if it's in the send or receive step because the error messages are identical and both return `nil`.

### DIFF
--- a/lib/kafka_ex/network_client.ex
+++ b/lib/kafka_ex/network_client.ex
@@ -5,10 +5,10 @@ defmodule KafkaEx.NetworkClient do
   def create_socket(host, port) do
     case :gen_tcp.connect(format_host(host), port, [:binary, {:packet, 4}]) do
       {:ok, socket} ->
-        Logger.log(:debug, "Succesfully connected to #{inspect(host)} on port #{inspect port}")
+        Logger.log(:debug, "Succesfully connected to broker #{inspect(host)}:#{inspect port}")
         socket
       _             ->
-        Logger.log(:error, "Could not connect to broker #{inspect(host)} on port #{inspect port}")
+        Logger.log(:error, "Could not connect to broker #{inspect(host)}:#{inspect port}")
         nil
     end
   end
@@ -23,7 +23,7 @@ defmodule KafkaEx.NetworkClient do
     case :gen_tcp.send(socket, data) do
       :ok -> :ok
       {_, reason} ->
-        Logger.log(:error, "Sending data to broker #{inspect broker.host} on port #{inspect broker.port} failed with #{inspect reason}")
+        Logger.log(:error, "Asynchronously sending data to broker #{inspect broker.host}:#{inspect broker.port} failed with #{inspect reason}")
         reason
     end
   end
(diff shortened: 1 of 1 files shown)


bjhaid:
Do you need an hex release for this?

dantswain:
No I can just point at the SHA.  Thanks! :+1:

dantswain:
@bjhaid Just wanted to say thanks again.  I was able to get this into production yesterday and it helped me debug issues overnight last night.

bjhaid:
Did you find the problem? If yes what was it?

dantswain:
We're still tracking that down, but I did learn some interesting things.  Sorry if not all of this brain dump is interesting to you, but it's helping me organize my own thoughts, too ;)

tl;dr version: 

I think we're experiencing some problems on our brokers that are causing timeouts.  A couple potential takeaways as they related to KafkaEx: 1) I'm finding it better in practice to use one KafkaEx worker per partition rather than sharing one across multiple partitions, 2) it may be worthwhile to expose KafkaEx.Supervisor's restart strategy parameters to the application config.

Much longer version:

* We're experiencing a timeout while reading from two of our brokers at certain times of day.
* It may be coincidence, or just another symptom of the underlying problem, but during those time frames we see a lot of ISR resizing happening on the brokers.  The ISR rebalancing is happening on different topics than the ones we are reading from with KafkaEx.
* This PR helped me isolate the timeout to the read portion of the transaction.  That's really valuable information!
* The timeouts were causing the KafkaEx worker to crash, which is fine.  The problem was that this caused a ripple effect and brought down enough linked processes that the whole VM was crashing.
* In my first iteration, I was sharing one KafkEx worker with six partition consumers.  In a gut instinct move, I changed that to one worker per partition.
  * At that point, I wasn't sure if the timeout was due to the broker being slow to respond or due to the consumer box experiencing resource contention.  I could be wrong, but my thinking was that sharing a KafkaEx worker process and a TCP/IP connection could be a source of contention, so using one worker per partition may relieve that.
  * Given what we observed last night (this issue is ongoing on our end), I'm mostly convinced that the timeouts are not due to resource contention on the consumer end.
  * The KafkaEx worker process crashing caused all of my consumer processes to crash because they were each linked to the worker process.  That number of process crashes exceeds Elixir's default restart strategy limits (3 restarts in 5 seconds iirc).  Using a separate KafkaEx worker process for each partition consumer process alleviates that because each KafkaEx worker process is only linked to a single consumer process.  I also upped the restart params in my supervisor (exposed it to config, actually).
  * It might be a good idea to expose the KafkaEx.Supervisor restart strategy parameters to the application config.  The number of allowed restarts may need to be tweaked as a function of the number of worker processes running on the user's VM.  I.e.,

```
# in config.exs
config :kafka_ex,
  max_restarts: 3,
  max_seconds: 5

# in the supervisor
@default_max_restarts 10
@default_max_seconds 60

def init(_args) do
  # ...
  supervisor_opts = [strategy: :simple_one_for_one, max_restarts: Application.get_env(:kafka_ex, :max_restarts, @default_max_restarts), max_seconds: Application.get_env(:kafka_ex, :max_seconds, @default_max_seconds)]
  # ...
end
```

This part is just my opinion; provided for the sake of discussion only - I'm not trying to say KafkaEx needs to do this.  Typically when I write library code in Elixir, I omit the supervisor and let the user (which is usually me, so take this with a grain of salt) provide their own supervisor for this reason among others (maybe the user wants a supervisor hierarchy for different parts of their application, etc.).  I would have something like `def KafkaEx.supervisor_spec(...)` that would return a  `Supervisor.Spec` that makes it relatively easy to work with.

bjhaid:
> It might be a good idea to expose the KafkaEx.Supervisor restart strategy parameters to the application config.

I recommended same in this issue https://github.com/kafkaex/kafka_ex/issues/65 though the OP in the issue decided a different direction for a fix of his problem. You can also bump the `sync_timeout` to match how long you think it takes before it times out.

You can also use a `monitor` instead of linking that way your are not completely taken down and you can restart the worker, or alternatively make the linked process supervised and let it do the starting/linking of kafka_ex workers as part of its init.

I would implement exposing the worker spec to config today.

PR #96, opened 2016-03-20
Make tests work under Erlang 7.3/Elixir 1.2.3

cdegroot:
I assume that the new output format of `:sys.get_state` is due to me upgrading to Erlang 7.3 - I guess we need a version check?


bjhaid:
Thanks for the PR, I have deleted the spec file itself, it relies the Supervisor State which is not documented and does not seem to be consistent across erlang versions

cdegroot:
:+1: I was mulling over these tests this morning and was about to propose something similarly rigorous :)

PR #97, opened 2016-03-20
Add 0.9 error codes

cdegroot:
Just some light typing work :-). Most of these can be returned from multiple 0.9 calls so it makes sense to add them all in one go.


bjhaid:
thanks! Just realized we have not been using this error_codes, I will make commit to ensure we use the error codes correctly when logging errors

Issue #86, opened 2016-03-14
Support for Group Membership - JoinGroup


bjhaid:
Piece meal from https://github.com/kafkaex/kafka_ex/issues/67

Specification can be found at:

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-JoinGroupRequest

cdegroot:
I'm gonna try my hand at it today, limited time though.

dantswain:
:+1: @cdegroot In case you haven't followed the conversation (see #67), the idea here was to split the overall task up into first implementing the individual structs and then working on the integration.

cdegroot:
Indeed, that's what I understood. Currently puzzling (with the Java client source next to it) what to put in the request ;-)

(off-topic: boy, is that Java client inefficient compared to what we can do in Elixir)

cdegroot:
https://github.com/kafkaex/kafka_ex/pull/98

bjhaid:
@cdegroot next time you could append

> closes kafkaex/kafkaex#98

to the commit message and github will auto magically tie the PR to the issue, see:
https://help.github.com/articles/closing-issues-via-commit-messages/

bjhaid:
thanks for the work and effort!

PR #98, opened 2016-03-20
Added JoinGroup request and response parsing

cdegroot:
I've taken a bit more verbose approach than the other code in the library, feel free to not agree with that code style :-)

### DIFF
--- a/lib/kafka_ex/protocol.ex
+++ b/lib/kafka_ex/protocol.ex
@@ -39,7 +39,7 @@ defmodule KafkaEx.Protocol do
   end
 
   defp api_key(:join_group) do
-    11
+    @join_group_request
   end
 
   def create_request(type, correlation_id, client_id) do
--- a/lib/kafka_ex/protocol.ex
+++ b/lib/kafka_ex/protocol.ex
@@ -6,6 +6,7 @@ defmodule KafkaEx.Protocol do
   @offset_commit_request     8
   @offset_fetch_request      9
   @consumer_metadata_request 10
+  @join_group_request        11
 
   @api_version  0
 
@@ -35,6 +36,10 @@ defmodule KafkaEx.Protocol do
 
   defp api_key(:consumer_metadata) do
     @consumer_metadata_request
+  end
+
+  defp api_key(:join_group) do
+    11
   end
 
   def create_request(type, correlation_id, client_id) do
--- a/lib/kafka_ex/protocol.ex
+++ b/lib/kafka_ex/protocol.ex
@@ -81,6 +81,7 @@ defmodule KafkaEx.Protocol do
     31 => :cluster_authorization_failed
   }
 
+  @spec error(integer) :: atom | integer
   def error(err_no) do
     case err_no do
       -1 -> :unknown_error
--- a/lib/kafka_ex/protocol/common.ex
+++ b/lib/kafka_ex/protocol/common.ex
@@ -1,0 +1,16 @@
+defmodule KafkaEx.Protocol.Common do
+  @moduledoc """
+  A collection of common request generation and response parsing functions for the
+  Kafka wire protocol.
+  """
+
+  @doc """
+  Generate the wire representation for a list of topics.
+  """
+  def topic_data([]), do: ""
+
+  def topic_data([topic|topics]) do
+    << byte_size(topic) :: 16-signed, topic :: binary >> <> topic_data(topics)
+  end
+
+end
--- a/lib/kafka_ex/protocol/join_group.ex
+++ b/lib/kafka_ex/protocol/join_group.ex
@@ -6,11 +6,11 @@ defmodule KafkaEx.Protocol.JoinGroup do
   @metadata_version 0
 
   defmodule Response do
-    defstruct error_code: 0, generation_id: 0, leader: "", members: []
-    @type t :: %Response{error_code: integer, generation_id: integer, leader: binary, members: list}
+    defstruct error_code: nil, generation_id: 0, leader: nil, members: []
+    @type t :: %Response{error_code: atom | integer, generation_id: integer, leader: binary, members: [binary]}
   end
 
-  @spec create_request(integer, binary, binary, binary, list, integer) :: binary
+  @spec create_request(integer, binary, binary, binary, [binary], integer) :: binary
   def create_request(correlation_id, client_id, member_id, group_name, topics, session_timeout) do
     KafkaEx.Protocol.create_request(:join_group, correlation_id, client_id) <>
       << byte_size(group_name) :: 16-signed, group_name :: binary,
--- a/lib/kafka_ex/protocol/join_group.ex
+++ b/lib/kafka_ex/protocol/join_group.ex
@@ -1,0 +1,43 @@
+defmodule KafkaEx.Protocol.JoinGroup do
+  import KafkaEx.Protocol.Common
+
+  @protocol_type "consumer"
+  @strategy_name "assign"
+  @metadata_version 0
+
+  defmodule Response do
+    defstruct error_code: 0, generation_id: 0, leader: "", members: []
+    @type t :: %Response{error_code: integer, generation_id: integer, leader: binary, members: list}
+  end
+
+  @spec create_request(integer, binary, binary, binary, list, integer) :: binary
+  def create_request(correlation_id, client_id, member_id, group_name, topics, session_timeout) do
+    KafkaEx.Protocol.create_request(:join_group, correlation_id, client_id) <>
+      << byte_size(group_name) :: 16-signed, group_name :: binary,
+         session_timeout :: 32-signed,
+         byte_size(member_id) :: 16-signed, member_id :: binary,
+         byte_size(@protocol_type) :: 16-signed, @protocol_type :: binary,
+         1 :: 32-signed, # We always have just one GroupProtocl
+         byte_size(@strategy_name) :: 16-signed, @strategy_name :: binary,
+         @metadata_version :: 16-signed,
+         length(topics) :: 32-signed, topic_data(topics) :: binary,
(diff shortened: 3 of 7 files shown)


lib/kafka_ex/protocol.ex
```diff
@@ -37,6 +38,10 @@ defmodule KafkaEx.Protocol do
     @consumer_metadata_request
   end
 
+  defp api_key(:join_group) do
+    11
```
bjhaid: I would use the module attribute `@join_group_request` here instead of the magic number 11

lib/kafka_ex/protocol/metadata.ex
```diff
@@ -56,12 +57,6 @@ defmodule KafkaEx.Protocol.Metadata do
     KafkaEx.Protocol.create_request(:metadata, correlation_id, client_id) <> << length(topics) :: 32-signed, topic_data(topics) :: binary >>
   end
 
-  defp topic_data([]), do: ""
-
-  defp topic_data([topic|topics]) do
-    << byte_size(topic) :: 16-signed, topic :: binary >> <> topic_data(topics)
-  end
-
```
bjhaid: :heart:

bjhaid:
I think the style choices are fine, I have been thinking of adding [credo](https://github.com/rrrene/credo) for style checking/linting, the code looks good to me, I'll leave this to @dantswain to merge

lib/kafka_ex/protocol.ex
cdegroot: Funny. I have numeric constants in all the other calls. I must be messing up with Git somewhere :-). Will fix it.

dantswain:
I'll take a look in a bit. Lots house work today :rage1:

 - Dan

lib/kafka_ex/protocol/join_group.ex
```diff
@@ -0,0 +1,43 @@
+defmodule KafkaEx.Protocol.JoinGroup do
+  import KafkaEx.Protocol.Common
+
+  @protocol_type "consumer"
+  @strategy_name "assign"
+  @metadata_version 0
+
+  defmodule Response do
+    defstruct error_code: 0, generation_id: 0, leader: "", members: []
```
dantswain: I'd say use default values of `nil` or `[]`.  `error_code: 0` might be a reasonable default if you're only setting an error if an error occurs.

lib/kafka_ex/protocol/join_group.ex
```diff
@@ -0,0 +1,43 @@
...
+    @type t :: %Response{error_code: integer, generation_id: integer, leader: binary, members: list}
```
dantswain: Can we narrow down the type for `members` a little more?  What is it a list of?

lib/kafka_ex/protocol/join_group.ex
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  @spec create_request(integer, binary, binary, binary, list, integer) :: binary
```
dantswain: Can we use `[binary]` for the type of `topics`?

lib/kafka_ex/protocol/metadata.ex
```diff
@@ -1,5 +1,6 @@
 defmodule KafkaEx.Protocol.Metadata do
   alias KafkaEx.Protocol
+  import KafkaEx.Protocol.Common
```
dantswain: I'm not a big fan of `import`, but if you guys disagree then I won't put up much of a fight.

test/integration/consumer_group_test.exs
```diff
@@ -211,4 +211,16 @@ defmodule KafkaEx.ConsumerGroup.Test do
 
     assert 0 >= offset_fetch_response_offset
   end
+
+  test "can join a consumer group" do
+    random_group = generate_random_string
+    random_topic = generate_random_string
+    KafkaEx.create_worker(:join_group, [uris: uris, consumer_group: random_group])
+    KafkaEx.produce(%Proto.Produce.Request{topic: random_topic, partition: 0, required_acks: 1, messages: [%Proto.Produce.Message{value: "hey"}]})
```
dantswain: Is the produce call necessary here?

lib/kafka_ex/protocol/common.ex
```diff
@@ -0,0 +1,16 @@
...
+  A collection of common request generation and response parsing functions for the
+  Kafka wire protocol.
+  """
+
+  @doc """
+  Generate the wire representation for a list of topics.
+  """
+  def topic_data([]), do: ""
+
+  def topic_data([topic|topics]) do
```
dantswain: Can you refactor this to be tail-call recursive?

dantswain: Oh, I see now that it wasn't tail-call recursive before.  As long as you're refactoring, it's probably a good idea to make that change.

bjhaid: @dantswain read http://erlang.org/doc/efficiency_guide/myths.html

dantswain: Huh.  I didn't know.  It still says tail-recursive tends to be faster on x86 hardware _shrugs_.

bjhaid: This is another educative read http://ferd.ca/erlang-s-tail-recursion-is-not-a-silver-bullet.html

test/integration/consumer_group_test.exs
cdegroot: It creates the topic. You can't setup consumer groups on non-existing topics.

lib/kafka_ex/protocol/metadata.ex
cdegroot: Random imports - meh. This is sort of a base class import, which I think is ok. However, an alias would work for me as well.

lib/kafka_ex/protocol/common.ex
cdegroot: Regardless of the above - let's not go overboard here, I'd say. You'll interact with a handful of topics at best, you'll never see a measurable difference even if TCO is a silver bullet. I'll eat my words and recat the function when a bug report gets filed, but without proof that someone needs to deal with a million topics, I think the cleaner code version should stay.

test/integration/consumer_group_test.exs
dantswain: I thought that might be the case, but you're creating a topic with a random name but then using `["foo", "bar"]` as the list of topics below.  Presumably the tests pass, so I'm wondering if the topics really need to be created?  Does the spec say this anywhere?

(If this does end up being the case, I'd say it's worth a comment in the test code because following all of the states that the broker/client needs to be in is pretty complex)

lib/kafka_ex/protocol/metadata.ex
dantswain: That argument makes sense.  As a matter of personal preference I still like to avoid `import` because it obfuscates the source of the imported functions.  FWIW my opinion is informed by trying to get other developers to adopt Elixir and observing what frustrates them.  

Either way, I'm just saying this for the sake of conversation, really.  I don't expect you to change it unless maybe @bjhaid has a strong opinion on the topic but I suspect he would have expressed it if he did ;)

lib/kafka_ex/protocol/common.ex
dantswain: So, also just for the sake of conversation, I think "cleaner" is subjective here.  For whatever reason, when I learned Erlang, I got "use tail call recursion" drilled into my head.  So I'm used to thinking that way and seeing something that's NOT tcr is jarring to me ;)  Which is just to say: there are different points of view on the topic.

dantswain:
Looks like the build failure was due to a network hiccup installing Kafka.  I restarted it.

cdegroot:
It's probably more robust (and certainly faster) if the kafka build gets cached. See e.g. https://github.com/PagerDuty/pd-travis-deps

dantswain:
Nice!

dantswain:
Looks like the host we're pulling Kafka from is down.  @cdegroot if you want to make a PR to integrate build caching, I'd look at it ASAP.

test/integration/consumer_group_test.exs
cdegroot: Heh. Good catch. Needless to say, "foo" and "bar" existing in my integration environment :-). Will double check and fix either way.

dantswain: Cool.  Let us know what you find.  They probably didn't exist in the travis env, so I'm suspecting the topics don't need to exist?  If they do need to exist and the test still passes, that's concerning :/

lib/kafka_ex/protocol/metadata.ex
bjhaid: I personally would not use `import` but I don't have a strong opinion against it

bjhaid:
I think you can only use travis cache on the travis enterprise offering

> The features described here are currently only available for private repositories on travis-ci.com and our new container-based infrastructure.

from https://docs.travis-ci.com/user/caching/

cdegroot:
@bjhaid I think the wording is not entirely clear, but let's test it before we go on shave this yak (looks like these scripts need some more work to be reasonable for both development and Travis)

dantswain:
@cdegroot that's why I was suggesting adding a little verbiage to the install script - so that it tells us explicitly whether or not it's using the cache.

cdegroot:
(we're in the wrong PR here to discuss Travis caching. Also - is there Slack/IRC presence for people working on this lib? I'm on both)

dantswain:
Fair enough ;)

@bjhaid and I were just discussing that (via email, ironically)

 - Dan

dantswain:
Closes #86 

We will have to revisit at some point re: error handling.  I think some of the errors that can get returned should be handled (rather than let-it-fail).

PR #99, opened 2016-03-21
Cache kafka installation to speed up travis builds.

cdegroot:
This will also remove the dependency on the Apache mirror being
available during build, as Kafka will only be installed once.

Note: to upgrade Kafka, it's from now on needed to junk the Travis
cache before pushing an install.sh pointing to a new version of
Kafka.

@dantswain let's see whether this does the trick.


install_kafka.sh
```diff
@@ -1,26 +1 @@
...
-bin/kafka-server-start.sh ../server.properties1 > /dev/null &
-echo $! > kafka1.pid
-sleep 2
-bin/kafka-server-start.sh ../server.properties2 > /dev/null &
-echo $! > kafka2.pid
-sleep 2
-bin/kafka-server-start.sh ../server.properties3 > /dev/null &
-echo $! > kafka3.pid
-sleep 5
+test -d ~/kafka || (mkdir ~/kafka; curl http://apache.mirrors.tds.net/kafka/0.9.0.1/kafka_2.11-0.9.0.1.tgz | tar -zxv -C ~/kafka --strip-components=1)
```
bjhaid: I think we also want to check that the directory is not empty, I also deleted the cache from the travis UI

install_kafka.sh
```diff
@@ -1,26 +1 @@
...
-bin/kafka-server-start.sh ../server.properties1 > /dev/null &
-echo $! > kafka1.pid
-sleep 2
-bin/kafka-server-start.sh ../server.properties2 > /dev/null &
-echo $! > kafka2.pid
-sleep 2
-bin/kafka-server-start.sh ../server.properties3 > /dev/null &
-echo $! > kafka3.pid
-sleep 5
+test -d ~/kafka/bin || (mkdir ~/kafka; curl http://apache.mirrors.tds.net/kafka/0.9.0.1/kafka_2.11-0.9.0.1.tgz | tar -zxv -C ~/kafka --strip-components=1)
```
dantswain: It would be good if this echoed something to stdout so that we can follow along in the build output.

Also I think it should test if we're in the travis env or not - I don't want `~/kafka` populated on my local machine ;)

start_kafka.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+BUILD_DIR=$(pwd)
```
dantswain: :+1:  I had done this locally and never actually committed it.  I have a couple local kafka clusters that I have to spin up/down depending on what I'm working on.

install_kafka.sh
cdegroot: It echoes a _ton_ of stuff to stdout. Have you clicked open the output for this step in the Travis build output? 

For the location - we can either work from the assumption that this is travis-only, or add bells and whistles. I'm working from the assumption that this is travis-only, I would never have run the old version that drops kafka in your build tree because I have a development environment with an actual Kafka running for work :)

bjhaid:
We probably want to do the things we are doing here only when one of the environment variables specified [here](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) is set, I will say we use `TRAVIS`, we also will need to modify `cleanup.sh` so it doesn't remove directories

install_kafka.sh
```diff
@@ -1,26 +1 @@
...
-sed -i.bak 's/@port@/9092/g' server.properties1
-cp server.properties server.properties2
-sed -i.bak 's/@broker_id@/2/g' server.properties2
-sed -i.bak 's/@port@/9093/g' server.properties2
-cp server.properties server.properties3
-sed -i.bak 's/@broker_id@/3/g' server.properties3
-sed -i.bak 's/@port@/9094/g' server.properties3
-mkdir kafka
-rm -rf /tmp/zookeeper
-rm -rf /tmp/kafka-logs*
```
bjhaid: I will want to keep removing this before we start kafka so we don't run into leftover data from previous runs

cdegroot:
Heh. Didn't notice cleanup.sh. Why is that there?

install_kafka.sh
cdegroot: Travis starts a fresh container for every run, which means a fresh /tmp. So the rm actions are a no-op.

bjhaid:
@cdegroot I use those scripts to install and cleanup kafka locally, they are around my workflow for kafka_ex

install_kafka.sh
dantswain: I mean specifically to let us know if the build is using the cached dependency or not.  I'm not a bash pro but it should just be something like

```
if [ -d "$KAFKA_DIR" ]; then
  echo "Using kafka at ${KAFKA_DIR}"
else
  echo "Installing kafka to ${KAFKA_DIR}"
fi
```

This is not travis-only.  I think @bjhaid and I both use this script.  I use it because it pulls the specific version of Kafka that we want to test against and because the cluster configuration here is different than my work-related cluster configuration.  I also have them configured to use different directories for kafka logs and zookeeper data, so one cluster cannot mess up the state of another and each one can be independently wiped clean.

scripts/start_kafka.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+source $(dirname $0)/kafka_base_dir.sh
+BUILD_DIR=$(pwd)/_build/kafka
+
+mkdir -p $BUILD_DIR
+
+for i in 1 2 3
+do
+  port=$(expr 9092 + ${i} - 1)
```
dantswain: Is there any reason this couldn't be combined with the loop below?

scripts/README.me
```diff
@@ -0,0 +1,8 @@
+This directory contains three scripts to:
```
bjhaid: was the extension of this intended to be `be` or was a mistake and should be `md`?

scripts/kafka_base_dir.sh
```diff
@@ -0,0 +1,6 @@
+if [ "$TRAVIS" == "true" ]
```
bjhaid: I prefer `[[` over `[` and I will also prepend the file with the shebang `#!/bin/bash`

bjhaid: also `[` doesn't work with `==` for equality see: http://mywiki.wooledge.org/BashFAQ/031

bjhaid:
Overall :heart:

scripts/start_kafka.sh
cdegroot: SRP in Bash. Both loops do one thing ;-). I thought about it, but this seemed cleaner in terms of flow through the script.

scripts/kafka_base_dir.sh
cdegroot: The hashbang is there. And the double-equals is a typo indeed. '[[' is hardly used as far as I can tell, would probably raise more eyebrows than be helpful in making this more readable.

scripts/README.me
cdegroot: Typo, thanks for spotting.

scripts/start_kafka.sh
bjhaid: I also prefer one loop over 2

bjhaid:
:ship: :it: except @cdegroot wants to make the 2 loops into one

cdegroot:
@cdegroot doesn't want that. Please merge ;-)

dantswain:
Will merge sometime later today.

dantswain:
@cdegroot I finally had a chance to pull this down and poke around a bit.  There's some places in the scripts where it assumes you're running it from the parent directory and won't work otherwise.  E.g.,

```
BUILD_DIR=$(pwd)/_build/kafka
```

and

```
base=$(pwd)/_build/kafka/dist
```

It should probably work relative to the scripts directory.  Something like this should work (see http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in - there are some gotchas but we are in a position where we control most of those)
```
scripts_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
build_dir=${scripts_dir}/../_build/kafka
```

dantswain:
Actually, you're already using `$(dirname $0)` a few places, so that's probably fine instead of the more complex version.

cdegroot:
I know and I'm fully aware of it. Note that the old scripts didn't do it either. I'm declaring this gold-plating - feel free to add it, but I'm gonna spend my time on the next 0.9 calls, which I think is a better use of my rare free time ;-)

dantswain:
Thank you, @cdegroot 
Could you please review #105 ?  I have tested it locally and on Travis.  The scripts now work when not called from the project root.

Issue #101, opened 2016-03-21
Test against multiple versions of elixir in travis


bjhaid:
Example:

https://github.com/elixircnx/comeonin/pull/11/files

dantswain:
I have a branch for this actually.

bjhaid:
Let's merge it :)

dantswain:
Closed by #102

PR #102, opened 2016-03-21
Test against multiple versions of Elixir and OTP

dantswain:
Re #101


dantswain:
@bjhaid A few thoughts I had on this: 

1) OTP 17.4 had a bug that would cause segfaults in the network client - I have already run afoul of this in production.  I thought it would clear up the segfaults we were seeing in the test suite if we specified 17.5 (where the segfault was fixed was fixed), but It doesn't :/  You may be right on it being related to docker.

2) OTP 18.3 is not yet supported by Travis

3) A potential downside of this is that we have a lot of spurious test failures and this will increase the probability that any given build will fail :/

bjhaid:
I sincerely will not pin against OTP versions, I will just list elixir versions, since elixir gradually drops support for older OTP versions

dantswain:
I have run afoul of that as well :/  Is there any way to tell Travis to use certain combinations?  The reason I wanted to test against different versions of OTP is that I have seen at least two bugs that were specifically related to which version of OTP I'm using :(

bjhaid:
https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix ?

bjhaid:
so something like:

```yaml
matrix:
  exclude:
    - elixir: 1.0.4
    - otp_version: 18.2.1
```

dantswain:
I like that.  Probably be a bit later today before I can get to it.

dantswain:
@bjhaid Check this out.  I did end up having to restart one of the builds.

bjhaid:
:+1: :shipit:

PR #106, opened 2016-03-25
 Add create_request and parse_response for SyncGroup

### LINKED ISSUE #87: Support for Group Membership - SyncGroup
Piece meal from https://github.com/kafkaex/kafka_ex/issues/67

Specification can be found at:

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-SyncGroupRequest

cdegroot:
Also added mix_test_watch, to automatically run unit tests during development.

WIP: Still needs an integration test

Closes #87

### DIFF
--- a/lib/kafka_ex/protocol/sync_group.ex
+++ b/lib/kafka_ex/protocol/sync_group.ex
@@ -1,0 +1,76 @@
+defmodule KafkaEx.Protocol.SyncGroup do
+  import KafkaEx.Protocol.Common
+
+  @member_assignment_version 0
+
+  defmodule Assignment do
+    defstruct topic: nil, partitions: []
+    @type t :: %Assignment{topic: binary, partitions: [integer]}
+  end
+
+  defmodule Response do
+    defstruct error_code: nil, assignments: []
+    @type t :: %Response{error_code: atom | integer, assignments: [Assignment.t] }
+  end
+
+  @spec create_request(integer, binary, binary, integer, binary, [{binary, [Assignment.t]}]) :: binary
+  def create_request(correlation_id, client_id, group_name, generation_id, member_id, assignments) do
+    KafkaEx.Protocol.create_request(:sync_group, correlation_id, client_id) <>
+    << byte_size(group_name) :: 16-signed, group_name :: binary,
+       generation_id :: 32-signed,
+       byte_size(member_id) :: 16-signed, member_id :: binary,
+       length(assignments) :: 32-signed, group_assignment_data(assignments, "") :: binary,
+       0 :: 32-signed
+    >>
+  end
+
+  @spec parse_response(binary) :: Response.t
+  def parse_response(<< _correlation_id :: 32-signed, error_code :: 16-signed,
+                        _member_assignment_len :: 32-signed,
+                        @member_assignment_version :: 16-signed,
+                        assignments_size :: 32-signed, rest :: binary >>) do
+    assignments = parse_assignments(assignments_size, rest, [])
+    %Response{error_code: KafkaEx.Protocol.error(error_code), assignments: assignments}
+  end
+
+  # Helper functions to create assignment data structure
+
+  defp group_assignment_data([], acc), do: acc
+  defp group_assignment_data([h|t], acc), do: group_assignment_data(t, acc <> member_assignment_data(h))
+
+  defp member_assignment_data({member_id, member_assignment}) do
+    assignment_bytes_for_member = <<
+       @member_assignment_version :: 16-signed,
+       length(member_assignment) :: 32-signed,
+       topic_assignment_data(member_assignment, "") :: binary >>
+    << byte_size(member_id) :: 16-signed, member_id :: binary,
+       byte_size(assignment_bytes_for_member) :: 32-signed,
+       assignment_bytes_for_member :: binary >>
+  end
+
+  defp topic_assignment_data([], acc), do: acc
+  defp topic_assignment_data([h|t], acc), do: topic_assignment_data(t, acc <> partition_assignment_data(h))
+
+  defp partition_assignment_data({topic_name, partition_ids}) do
+    << byte_size(topic_name) :: 16-signed, topic_name :: binary,
+       length(partition_ids) :: 32-signed, partition_id_data(partition_ids, "") :: binary >>
+  end
+
+  defp partition_id_data([], acc), do: acc
+  defp partition_id_data([h|t], acc), do: partition_id_data(t, acc <> << h :: 32-signed >>)
+
+  # Helper functions to parse assignments
+
+  defp parse_assignments(0, _rest, assignments), do: assignments
+  defp parse_assignments(size, << topic_len :: 16-signed, topic :: size(topic_len)-binary,
+                                  partition_len :: 32-signed,
+                                  rest :: binary >>, assignments) do
+    {partitions, rest} = parse_partitions(partition_len, rest, [])
+    parse_assignments(size - 1, rest, [{topic, partitions}|assignments])
+  end
+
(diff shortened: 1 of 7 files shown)


lib/kafka_ex/protocol/sync_group.ex
```diff
@@ -0,0 +1,66 @@
+defmodule KafkaEx.Protocol.SyncGroup do
+  import KafkaEx.Protocol.Common
+
+  @member_assignment_version 0
+
+  defmodule Response do
+    defstruct error_code: nil, assignments: []
+    @type t :: %Response{error_code: atom | integer, assignments: [{binary, [integer]}]}
```
dantswain: Can `error_code` be a generic atom or can we narrow it down to `nil | integer`?

Also, it might help to create a type for `assignment`.  I might even go so far as to create types for the `binary` and `integer` components of an assignment - i.e., what _are_ they?  I think it helps when trying to digest the code at a glance.  It can also generate helpful dialyzer warnings.

mix.exs
```diff
@@ -22,6 +22,7 @@ defmodule KafkaEx.Mixfile do
       {:earmark, "~> 0.1", only: :dev},
       {:dialyze, "~> 0.2", only: :dev},
       {:ex_doc, "~> 0.7", only: :dev},
+      {:mix_test_watch, "~> 0.2", only: :dev},
```
dantswain: I'd prefer to keep pull requests limited in scope to one functional change.

dantswain:
Looks good so far other than my minor nits.

lib/kafka_ex/protocol/sync_group.ex
cdegroot: Well, given that error codes are ran through `KafkaEx.Protocol.error` which has a `atom | integer` signature, there's not much to do here (or anywhere else, I'd say. I wonder why dialyzer let it through).

+1 on the type aliases. Here and everwhere else btw ;-)

lib/kafka_ex/protocol/sync_group.ex
```diff
@@ -0,0 +1,76 @@
...
+       generation_id :: 32-signed,
+       byte_size(member_id) :: 16-signed, member_id :: binary,
+       length(assignments) :: 32-signed, group_assignment_data(assignments, "") :: binary,
+       0 :: 32-signed
+    >>
+  end
+
+  @spec parse_response(binary) :: Response.t
+  def parse_response(<< _correlation_id :: 32-signed, error_code :: 16-signed,
+                        _member_assignment_len :: 32-signed,
```
dantswain: Something doesn't line up here.  The guide says the format of the response is

```
SyncGroupResponse => ErrorCode MemberAssignment
  ErrorCode => int16
  MemberAssignment => bytes
```

And, afaict, MemberAssignment is

```
MemberAssignment => Version PartitionAssignment
  Version => int16
  PartitionAssignment => [Topic [Partition]]
    Topic => string
    Partition => int32
  UserData => bytes
```

So I'm not seeing where `_member_assign_len :: 32-signed` fits in.

Also, what is `version` here?  Is it a protocol version or a version of the group structure?  The protocol is pegged at 0, but the version of the group structure will not be.  I can try to dig through the docs more tomorrow for details.

mix.lock
```diff
@@ -1,4 +1,6 @@
 %{"dialyze": {:hex, :dialyze, "0.2.1"},
   "earmark": {:hex, :earmark, "0.1.17"},
   "ex_doc": {:hex, :ex_doc, "0.9.0"},
+  "fs": {:hex, :fs, "0.9.2"},
```
dantswain: `mix deps.clean --unused --unlock` should get rid of these

lib/kafka_ex/protocol/sync_group.ex
bjhaid: `_member_assign_len` is the number of bytes contained in `MemberAssignment`, maybe we should really use it and not throw it away, it's used it in the java client see: https://github.com/apache/kafka/blob/d9bf55171ba41962edc05cdf106f540cf94a2c3e/clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java#L287-L289

bjhaid: also it's [documented](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ProtocolPrimitiveTypes):

> Variable Length Primitives
bytes, string - These types consist of a signed integer giving a length N followed by N bytes of content. A length of -1 indicates null. string uses an int16 for its size, and bytes uses an int32.

bjhaid: `version` is sort of tricky it is not clearly documented anywhere neither does the [java client set it](https://github.com/apache/kafka/blob/d9bf55171ba41962edc05cdf106f540cf94a2c3e/clients/src/main/java/org/apache/kafka/common/requests/SyncGroupRequest.java) so 0 is fine for now, we probably should ask questions from the kafka mailing list what use `Version` is for

cdegroot: Background: MemberAssignment is, from the broker's perspective, opaque. This allows clients to coordinate together - they join the group, one gets made leader by the broker and gets to do the member assignments, and then the leader sends whatever data the group has agreed upon to the broker to send around. Therefore, MemberAssignment is primary an opaque bytestring to the broker, hence the need to send it length+data. Then, the "consumer" protocol as implemented in the Java client goes on to interpret the MemberAssignment structure as `[{topic, [partition]}]` which my code supports making it theoretically viable to mix Java and kafka_ex clients (if you ever want to do that ;-)). 

On using the length: I had most of the unpacking code done when I found out about the above. It turns out that when you know the contents of the structure, you don't need the length. The Java client has a completely different approach to (de)serialization through the generic Struct class and metadata descriptions, I don't think there's much for us to pick up.

mix.lock
cdegroot: Will clean up

lib/kafka_ex/protocol/sync_group.ex
bjhaid: > Background: MemberAssignment is, from the broker's perspective, opaque. This allows clients to coordinate together - they join the group, one gets made leader by the broker and gets to do the member assignments, and then the leader sends whatever data the group has agreed upon to the broker to send around. Therefore, MemberAssignment is primary an opaque bytestring to the broker, hence the need to send it length+data. Then, the "consumer" protocol as implemented in the Java client goes on to interpret the MemberAssignment structure as [{topic, [partition]}] which my code supports making it theoretically viable to mix Java and kafka_ex clients (if you ever want to do that ;-)).

I sort of think this is not controversial, the part that is controversial is the API docs has `Version` as an attribute on `MemberAssignment` but there is no documentation on what `Version` means nor is it implemented in the reference implementation, which is why I agree that leaving it as `NULL` (0) is fine for now.

> On using the length: I had most of the unpacking code done when I found out about the above. It turns out that when you know the contents of the structure, you don't need the length. The Java client has a completely different approach to (de)serialization through the generic Struct class and metadata descriptions, I don't think there's much for us to pick up.

While I agree with this, this approach is inconsistent with what we do in other places, e.g:

https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/protocol/fetch.ex#L76
https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/protocol/fetch.ex#L83

cdegroot: This would mean we need extra indirections in the parsing here, further complicating the parsing code. As long as we don't have error-handling for short packets anywhere else, I'm not sure that this complication warrants the work.

Issue #109, opened 2016-04-08
When fetching: `auto_commit` should default to false if `:no_consumer_group` is set


sdball:
If I have `consumer_group: :no_consumer_group` then I get errors when fetching from a topic:

```elixir
KafkaEx.fetch("foo", 0, offset: 0)
** (exit) exited in: GenServer.call(KafkaEx.Server, {:fetch, "foo", 0, 0, 10, 1, 1000000, true}, 5000)
    ** (EXIT) an exception was raised:
        ** (MatchError) no match of right hand side value: false
            (kafka_ex) lib/kafka_ex/server.ex:91: KafkaEx.Server.handle_call/3
            (stdlib) gen_server.erl:629: :gen_server.try_handle_call/4
            (stdlib) gen_server.erl:661: :gen_server.handle_msg/5
            (stdlib) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
    (elixir) lib/gen_server.ex:564: GenServer.call/3

** (MatchError) no match of right hand side value: false
    (kafka_ex) lib/kafka_ex/server.ex:91: KafkaEx.Server.handle_call/3
    (stdlib) gen_server.erl:629: :gen_server.try_handle_call/4
    (stdlib) gen_server.erl:661: :gen_server.handle_msg/5
    (stdlib) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
Last message: {:fetch, "foo", 0, 0, 10, 1, 1000000, true}
[...]

PR #110, opened 2016-04-10
Update documentation wrt auto_commit as per #109

dantswain:
Offsets can not be committed when a consumer group is not set.  We are discussing a longer-term fix for this.  Updating the documentation in the short term seems like a good idea.

@sdball do you think this would have helped you avoid the problems you reported in #109?


sdball:
Yes indeed, thanks!

PR #122, opened 2016-06-11
fixes README for auto_commit and streaming

blackfist:


README.md
```diff
@@ -177,16 +177,16 @@ iex> KafkaEx.produce("foo", 0, "hey") # where "foo" is the topic and "hey" is th
...
 iex> KafkaEx.create_worker(:stream, [uris: [{"localhost", 9092}]])
 {:ok, #PID<0.196.0>}
-iex> KafkaEx.produce("foo", 0, "hey", :stream)
+iex> KafkaEx.produce("foo", 0, "hey", worker_name: :stream)
 :ok
-iex> KafkaEx.produce("foo", 0, "hi", :stream)
+iex> KafkaEx.produce("foo", 0, "hi", worker_name: :stream)
 :ok
-iex> KafkaEx.stream("foo", 0) |> Enum.take(2)
+iex> KafkaEx.stream("foo", 0, offset: 0, auto_commit: false) |> Enum.take(2)
```
bjhaid: You should remove `auto_commit: false` as we don't want to provide an example that switches off `auto-commit` or you provide a second example for kafka < 0.8.2 and leave the original example as it is. Otherwise this is fine, thanks for the PR, ❤️

blackfist:
OK, I made that change to the PR and squashed it down to one commit.

bjhaid:
thanks ❤️

PR #124, opened 2016-06-13
moves one test into an integration test

### LINKED ISSUE #123: Failing test for disable_default_worker
When you run `mix test --no-start` one of the tests fails.

```
  1) test Setting disable_default_worker to true removes the KafkaEx.Server worker (KafkaExTest)
     test/kafka_ex_test.exs:4
     ** (MatchError) no match of right hand side value: {:error, {:not_started, :kafka_ex}}
     stacktrace:
       test/kafka_ex_test.exs:6
```

The reason for the failing test is that the test tries to run `Application.stop(:kafka_ex)` but since we run `mix test` with `--no-start` there is no `:kafka_ex` application to stop. I talked to @bjhaid in the kafka_ex chat room and he's cool with moving that test in with the other integration tests.

I'll try to send a PR today.

blackfist:
Closes #123

### DIFF
--- a/test/integration/integration_async_test.exs
+++ b/test/integration/integration_async_test.exs
@@ -1,7 +1,8 @@
-defmodule KafkaExTest do
+defmodule KafkaEx.Integration.Async.Test do
   use ExUnit.Case, async: false
 
-  @tag :integration
+  @moduletag :integration
+
   test "Setting disable_default_worker to true removes the KafkaEx.Server worker" do
     # stop the application, disable the default worker, restart the application
     :ok = Application.stop(:kafka_ex)


test/integration/integration_async_test.exs
```diff
@@ -1,6 +1,8 @@
-defmodule KafkaExTest do
+defmodule KafkaEx.Integration.Async.Test do
```
bjhaid: I prefer we keep `KafkaExTest` other than that I am 👍  to this, thanks!

blackfist:
Ok, I put the file back and made the module with nothing in it for now.

blackfist:
Alright, based on our conversation yesterday in Slack I think I've got this where we want it. The test is back in the KafkaExTest module and I added the `:integration` tag to that one test.

Issue #119, opened 2016-05-27
kafka_ex server refactor


bjhaid:
### Motivation

The consumer_group API (storing of offsets) in Kafka's API has changed continuously in versions on Kafka since the Kafka core team decided to stop storing offsets in zookeeper. This change in API leaves a client implementer with either adding new features and not supporting older versions of Kafka, or codifying backwards compatibility and not breaking the client for folks who can't afford an upgrade.

I have decided that the later option is better allowing folks who can't afford a Kafka upgrade to continuously enjoy using newer versions of kafka_ex (which can include bug fixes for parts that affects them), while folks who run the latest version of Kafka get the latest features we implement.

### Proposed Solution

Create a server for each version (or range of versions with similar API) removing `consumer_group` related code completely from server for kafka < 0.8.2, move the current `consumer_group` logic into `0.8.{2,3}` server and open the way for >= 0.9 implementations (via their own servers), the user supplies the version in configuration at worker creation we spawn a server matching their config. When they attempt an operation not supported for the version of kafka they configured we should raise an exception and crash the worker.

There is a WIP branch which should serve has a base for the proposed solution at https://github.com/kafkaex/kafka_ex/commit/ee89d667ac1df1203b26b7ee1862fd6989a7373d

PR #125, opened 2016-06-14
Use behaviors for servers

bjhaid:
WIP to close https://github.com/kafkaex/kafka_ex/issues/119

@dantswain please review. Thanks!


blackfist:
I know you didn't ask me to review it, but I did anyway 😉 

I like where this is going. I admit that it took me a bit to wrap my head around the code, but after I studied it for 20 minutes it started to come together. I think this is some up-front work that seems a bit complicated, but it's going to make it really easy to implement the next server api.

One question I had though, it looks like we're grabbing the server version from the Application.env. I didn't see a way to specify the server version if you're using `create_worker` or a way to override that. Does that mean that this library will not be able to support the use case of having multiple workers connecting to different versions of Kafka or did I just miss that bit of code?

bjhaid:
> Does that mean that this library will not be able to support the use case of having multiple workers connecting to different versions of Kafka or did I just miss that bit of code?

Yes, erlang supervisors are instantiated with a worker spec which it will use to create workers in the future (in the case on a `single_one_for_one`). I don't think there is any documented way to change the worker spec, there are ways you could manipulate the supervisor state, but that's not documented and can break anytime in the future.

We should probably document this behavior, that you can't switch workers in between.

Thanks for the review, I really appreciate it

dantswain:
We might want to provide functions that return supervisor specs so that the user can manage the workers themselves (I prefer this, personally... what if I want different workers under different supervisors?).  That would allow for connecting to multiple broker versions.

lib/kafka_ex.ex
```diff
@@ -39,7 +39,7 @@ defmodule KafkaEx do
   def create_worker(name, worker_init \\ []) do
     case build_worker_options(worker_init) do
       {:ok, worker_init} ->
-        Supervisor.start_child(KafkaEx.Supervisor, [worker_init, name])
+        Supervisor.start_child(KafkaEx.Supervisor, [server, worker_init, name])
```
dantswain: I think `server` maybe should have a different name.  Even knowing ahead of time what your intention was, it threw me for a loop looking for where the var `server` is defined.  `server_from_config` or `default_server` or something?  Another approach I've used that I like is to introduce a `Config` module and move all of the `get_env` stuff to that.  That way when you see `KafkaEx.Config.default_server_impl` it's pretty obvious where that value is coming from.

lib/kafka_ex.ex
```diff
@@ -137,7 +137,7 @@ defmodule KafkaEx do
 
   Optional arguments(KeywordList)
   - offset: When supplied the fetch would start from this offset, otherwise would start from the last committed offset of the consumer_group the worker belongs to. For Kafka < 0.8.2 you should explicitly specify this.
-  - worker_name: the worker we want to run this fetch request through. Default is KafkaEx.Server
+  - worker_name: the worker we want to run this fetch request through. Default is KafkaEx.server
```
dantswain: This probably warrants some explanation.  Even if it's just "See [some other part of the doc where we explain what KafkaEx.server is]"

dantswain:
One overall comment - we could probably resolve a lot of the ambiguity in the code if we use a different name to register the default worker.  There's no reason it has to be called `KafkaEx.Server` - any atom will suffice.  That would limit the "things that have to change per broker version" to just the server initialization.

lib/kafka_ex/default_server.ex
```diff
@@ -0,0 +1,309 @@
+defmodule KafkaEx.DefaultServer do
```
dantswain: Is this the 0.8 implementation?  Maybe just call it that?

dantswain:
I think the approach looks good.  I have a little bit mixed opinions on how you're handling the behavior - you've basically reimplemented GenServer (intentionally) rather than implementing on _top_ of GenServer.  The way you've done it is very powerful and clean, but duplicates some implementation from GenServer and wouldn't pick up improvements to GenServer from updates to the Elixir standard lib.

If you're not sure what I mean by implementing a behavior on top of GenServer, hit me on slack and I can dig up an example.

lib/kafka_ex.ex
bjhaid: I like the idea of introducing a `Config` module seems neat

lib/kafka_ex.ex
bjhaid: Yeah! this is me being lazy

lib/kafka_ex/default_server.ex
bjhaid: No there is a [0.8](https://github.com/kafkaex/kafka_ex/pull/125/files?diff=split#diff-3) server

bjhaid:
@dantswain thanks for the review I will definitely hit you up on slack

bjhaid:
@dantswain please take a second pass through this PR, I want to merge it ASAP

lib/kafka_ex/config.ex
```diff
@@ -0,0 +1,17 @@
...
+    :kafka_ex
+  end
+
+  def server_impl do
+    :kafka_ex
+      |> Application.get_env(:kafka_version, :default)
+      |> server
+  end
+
+  defp server("0.8.0"), do: KafkaEx.Server0P8P0
```
dantswain: Not sure this belongs here, it's not really "configuration".  Not a huge deal, though.

lib/kafka_ex/default_server.ex
dantswain: Which version is this for?  I guess what I'm saying is why not just call it KafkaEx.Impl0p9 or whatever version it implements?

dantswain:
@bjhaid I think I need to pull this down and poke around to really understand what's going on.  Unfortunately I'm pretty overworked right now :/  If you want to merge, go ahead and we could always refactor later?

lib/kafka_ex/default_server.ex
bjhaid: This has 0.8.2 and 0.9.0 things, until I peel them out, I want to keep it named default, it probably will end up being 0.9.0

bjhaid:
> I think I need to pull this down and poke around to really understand what's going on. Unfortunately I'm pretty overworked right now :/ If you want to merge, go ahead and we could always refactor later?

Cool by me, will do so, thanks!

PR #126, opened 2016-06-20
Make ExUnit capture log output

eproxus:
This cleans up the test output a bit by hiding logs unless tests fail. I haven't been able to clean up the initial log output which comes from the fact that Mix by default starts the application when running tests.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -22,3 +22,8 @@ config :kafka_ex,
   max_restarts: 10,
   # Supervision max_seconds -  the time frame in which :max_restarts applies
   max_seconds: 60
+
+env_config = Path.expand("#{Mix.env}.exs", __DIR__)
+if File.exists?(env_config) do
+  import_config(env_config)
+end
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,0 +1,3 @@
+use Mix.Config
+
+config :ex_unit, capture_log: true


bjhaid:
you can actually have `config :logger, level: :warn` in a test.exs file and include a `
import_config "#{Mix.env}.exs"` at the bottom of https://github.com/kafkaex/kafka_ex/blob/master/config/config.exs and that should do the trick

bjhaid:
that will only log on error or warning

eproxus:
Yes, but then you won't see the debug logging which might help you in debugging failed test cases.

eproxus:
And you will still get the error logging produced by test cases.

bjhaid:
how do you intend using the attribute `capture_log`???

eproxus:
See the *Log Capture* section here: http://elixir-lang.org/docs/stable/ex_unit/ExUnit.Case.html

I'd also be open for setting it in the test configuration globally instead, via `config :ex_unit, capture_log: true`.

bjhaid:
> I'd also be open for setting it in the test configuration globally instead, via config :ex_unit, capture_log: true

Either is fine by me, thanks for enlightening me once you are ready I will merge this

eproxus:
Last commit now adds `config/test.exs` which disables logging globally for all tests.

bjhaid:
Cool, I will merge sometimes today, thanks!

Issue #121, opened 2016-06-09
Consuming unstable when broker in cluster goes down


leifg:
I am currently evaluating kafka for internal purposes.

I set up a kafka cluster with 3 nodes (kafka.1, kafka.2 and kafka.3). the topic (name: "topic") I am subscribing to has 3 partitions and is replicated over 3 nodes:

```bash
$ /opt/kafka/bin/kafka-topics.sh --describe --zookeeper $KAFKA_ZOOKEEPER_CONNECT --topic topic
Topic:topic  PartitionCount:3        ReplicationFactor:3     Configs:
        Topic: topic Partition: 0    Leader: 3       Replicas: 3,1,2 Isr: 2,3,1
        Topic: topic Partition: 1    Leader: 1       Replicas: 1,2,3 Isr: 2,3,1
        Topic: topic Partition: 2    Leader: 2       Replicas: 2,3,1 Isr: 2,3,1
```

It's all great I am subscribing to the topic in a very simple way (endless loop, fetch all the time):

```elixir
defp fetch_loop(topic, partition, offset) do
  new_offset = KafkaEx.fetch(topic, partition, offset: offset) |> handle_batch
  persist_offset(new_offset)
  fetch_loop(topic, partition, new_offset)
end
```

This works fine.

But now I also wanted to test what happens if one of the nodes goes down. In theory that should not matter as the cluster will redistribute the leaders.

However my script will crash and display this message:

```elixir
16:02:39.053 [error] Sending data to broker "kafka.3", 9094 failed with :timeout

16:02:39.054 [debug] Shutting down worker KafkaEx.Server

16:02:39.081 [debug] Succesfully connected to "kafka" on port 9092
** (exit) exited in: GenServer.call(KafkaEx.Server, {:fetch, "topic", 2, 63194, 10, 1, 1000000, true}, 5000)
    ** (EXIT) an exception was raised:
        ** (FunctionClauseError) no function clause matching in KafkaEx.Protocol.Fetch.parse_response/1
            (kafka_ex) lib/kafka_ex/protocol/fetch.ex:18: KafkaEx.Protocol.Fetch.parse_response(nil)
            (kafka_ex) lib/kafka_ex/server.ex:324: KafkaEx.Server.fetch/8
            (kafka_ex) lib/kafka_ex/server.ex:92: KafkaEx.Server.handle_call/3
            (stdlib) gen_server.erl:629: :gen_server.try_handle_call/4
[...]

PR #127, opened 2016-06-20
Connection errors

eproxus:
*(Note: this builds on PR #126 for now, but it is optional)*

Here's a first try at introducing connection errors (partly to solve #121 and to just give developers a nicer API than just function clauses and nil errors to work with). Comments and concerns are welcome!

The implementation will raise `KafkaEx.Network.Error` on connection related problems (connect, send and receive). These can be rescued by the user and they can either retry or crash or decide to handle the error in any other way.

The following issues are still open:

1. What to do in the worker? What we need is what is now experimentally implemented for `KafkaEx.fetch/2` and that is that the worker sees the error, logs it, but doesn't crash. The error is then raised to the caller to deal with. It allows us to see that the worker failed to connect/send/receive to the broker and control it from our application (try again at a later time for example, or crash). Is this a sound model, that should be extended for the other API calls?
2. The test *"Creates a worker even when the one of the provided brokers is not available"* (`test/integration/integration_test.exs:21`) is failing because the connect fails. I'm not sure how much sense it makes for the worker to start when a connection is not possible? Does the server process try to reconnect? If not, I would let this blow up and not start a worker anyway.
3. The test *"publish/fetch handles a 10kb message snappy compressed"* fails. I don't know why, help is appreciated. This also fails on `master`!
4. Some tests in the `ConsumerGroup.Test` suite are failing. I don't know why, help is appreciated. These also fail on `master`!
5. Typespecs and running Dialyzer. `master` already has tons of Dialyzer warnings (31), how to proceed here?
6. Document the errors.


eproxus:
Oh, and as far as I could tell, the active mode on the socket was never really used? Please correct me if I'm wrong.

bjhaid:
You are sort of correct, we use it to dump all packets we don't care about on the floor (this is probably not necessary but a safe measure)

eproxus:
Another discussion that came up for us was the worker/server in the driver in general. Our problem is, that we have our own workers anyway for each partition in Kafka that uses their own KafkaEx worker. We would like to be able to use a more low level API in KafkaEx and avoid having KafkaEx spawn any processes at all. In that case, getting a connection error could be handled in the application code instead of having to be shipped through the KafkaEx server process all the time.

dantswain:
Can I ask why you would like to avoid spawning a process for the worker?  There shouldn't really be any appreciable overhead, and if you have 1-1 correspondence of your handlers and the KafkaEx workers, there shouldn't be any mailbox contention.  The plus side is that your handler and the KafkaEx worker don't have to share any state or logic.

I _do_ kind of wish that KafkaEx had a simpler way to let me manage the supervision of KafkaEx workers  as opposed to always starting them under the KafkaEx supervision tree.  So I'm sort of wondering if that's really what you are aiming for?

eproxus:
2 main reasons: overhead and error handling.

*Overhead:* There is overhead in that every message from Kafka is copied from the KafkaEx worker to our worker. There is no contention, correct, but the copying is expensive when you want to consume at high speed (we're currently consuming at 90 Mb/s on a 1 Gb link) and are CPU bound.

*Error handling:* The use case mentioned above is a good illustration. If a connection error occurs in the Kafka driver, we'd like to have control over that, deciding when and how to reconnect (if at all). If the connection (socket) is wrapped in a KafkaEx process, we lose that control and are bound by what the KafkaEx API currently exposes. That is, a crashing worker process that has to be managed. Restarting it from the supervisor might not always be desirable, since we might want to do exponential back-off (not supported by Erlang's supervisor). If we could use the connection itself we could decide how our worker used that connection, and implement our own back-off as desired.

eproxus:
@dantswain Yes, that as well. I'm not opposed to KafkaEx having some support for running a process, wether that be under it's own supervisor or just providing the GenServer implementation and a worker spec. In the end, I would like to be in control. :smile:

dantswain:
Re: Overhead - It seems unlikely to me that message copying is your bottleneck, but maybe you've profiled it?

I may put together something that would make it easier to create workers under your own supervision tree as I do think that would be a pretty useful feature.  It really just amounts to creating a function that returns a supervisor spec for you to deal with as you see fit (and properly documenting it).  @bjhaid thoughts on that?

eproxus:
For us, the workers under the KafkaEx supervisor are quite okay. We would just like to bypass them and maintain the Kafka connection ourselves.

I would propose a low-level "public" `KafkaEx.Connection` (or similar) functional API that works on the data structure (containing necessary metadata, socket etc.). This API should then also be used internally to implement workers (wether they live under the KafkaEx supervisor or are just part of a spec).

eproxus:
@dantswain I admit the copying is not the major part of performance for us, so it is more about the error handling. Although I don't like unnecessary inefficiencies when I see them. 😄

bjhaid:
>>> I may put together something that would make it easier to create workers under your own supervision tree as I do think that would be a pretty useful feature. It really just amounts to creating a function that returns a supervisor spec for you to deal with as you see fit (and properly documenting it). @bjhaid thoughts on that?

@dantswain 👍

bjhaid:
> I would propose a low-level "public" KafkaEx.Connection (or similar) functional API that works on the data structure (containing necessary metadata, socket etc.). This API should then also be used internally to implement workers (wether they live under the KafkaEx supervisor or are just part of a spec).

@eproxus this is a lot of work, I also can't figure out how this API will be used (a general use case) or what it will look like

eproxus:
@bjhaid I agree that it is a more substantial change, yes. I think in practice it doesn't have to be that big though. It is already in place in some places, like https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/server.ex#L271 for example. Also, the broker data structure that already exists (and contains the host, port and eventually socket) could be the functional carrier of the state.

eproxus:
An alternative way to view it would be, how do we solve the connection issues in a smooth way? Currently, our problem is that a connection issue in the KafkaEx worker will crash the worker and restart it via a supervisor. Just restarting it infinitely would lead to a too high CPU load and would spam the Kafka cluster. Having it in our own supervisor with help of the work of @dantswain would give us some more control, but in the end we would have to work around the normal supervisor logic anyway to achieve some kind of cooldown/backoff logic.

eproxus:
And yet another point (sorry for the spam!): We are using this in production at Wooga, so if we decide we really need this feature we would be happy to spend some development time on this.

bjhaid:
@eproxus Can we please take this discussion somewhere else, slack or a completely new issue, this PR is yet to be addressed and the discussion here has nothing to do with the PR

lib/kafka_ex/network/client.ex
```diff
@@ -0,0 +1,67 @@
+defmodule KafkaEx.Network.Client do
+  require Logger
+
+  alias KafkaEx.Network.{TCP, Error}
```
bjhaid: This is not supported in elixir < 1.2 and we have to support elixir > 1.0 please use the longer form

lib/kafka_ex.ex
```diff
@@ -167,10 +167,15 @@ defmodule KafkaEx do
 
     retrieved_offset = current_offset(supplied_offset, partition, topic, worker_name)
 
-    GenServer.call(worker_name, {
+    case GenServer.call(worker_name, {
```
bjhaid: I will extract this into an helper function as this will be a common pattern

lib/kafka_ex/network/tcp.ex
```diff
@@ -0,0 +1,19 @@
+defmodule KafkaEx.Network.TCP do
```
bjhaid: Is there any reason why you are creating this wrapper module?

lib/kafka_ex/network/client.ex
eproxus: Good point.

lib/kafka_ex.ex
eproxus: Yeah, that makes sense. I can look into it.

lib/kafka_ex/network/tcp.ex
eproxus: Dependency injection, for testing purposes. If you look at https://github.com/wooga/kafka_ex/blob/31cd90697729fbe1f6eba8ef86a2eff512dd6232/test/network_client_test.exs you can see it being replace with a mock module that doesn't interact with real sockets at all.

bjhaid:
See my answers in line:

> What to do in the worker? What we need is what is now experimentally implemented for KafkaEx.fetch/2 and that is that the worker sees the error, logs it, but doesn't crash. The error is then raised to the caller to deal with. It allows us to see that the worker failed to connect/send/receive to the broker and control it from our application (try again at a later time for example, or crash). Is this a sound model, that should be extended for the other API calls?

So there are a couple of concerns:
- The kafka application (I mean application in erlang terms here) requires connection to kafka on start up by default (though you can bypass this via `disable_default_worker: true` config option), this error you cannot catch, since it happens before your code is loaded.
- If kafka really crashes hard you want the workers processing the messages to crash and not pretend all is well, I willing to be convinced about alternatives.

In all I agree it's better to have saner exception/error messages than what we currently have.

> The test "Creates a worker even when the one of the provided brokers is not available" (test/integration/integration_test.exs:21) is failing because the connect fails. I'm not sure how much sense it makes for the worker to start when a connection is not possible? Does the server process try to reconnect? If not, I would let this blow up and not start a worker anyway.

So this test underlines how kafka/zookeeper works, we need only connection to one broker to determine the state of the cluster so we go through each of the brokers supplied and attempt connection to them and only keep state of the ones we could successfully connect to, we also rebalance both async and sync based on the hosts provided by kafka/zookeeper this allows the user to add brokers to their cluster seamlessly and also remove a broker in a controlled manner without having to update their application configuration or restart their application.

> The test "publish/fetch handles a 10kb message snappy compressed" fails. I don't know why, help is appreciated. This also fails on master!
I can't reproduce this failure, also does not fail on travis, I will be happy to help debug this.

> Some tests in the ConsumerGroup.Test suite are failing. I don't know why, help is appreciated. These also fail on master!
Some ConsumerGroup.Test will fail on first run because we create this consumer_groups on the fly and the kafka cluster takes some time to sync this information, so you need to run that spec twice (distributed systems are tough)

> Typespecs and running Dialyzer. master already has tons of Dialyzer warnings (31), how to proceed here?

Most of these dialyzer errors are:
- `Expression produces a value of type 'ok' | {'error',_}, but this value is unmatched` which expects something like `:ok = Logger.log...` which is sort of dumb, but we can do it to make dialyzer happy or remove the `--unmatched-returns` flag to `mix dialyze`
- `Type specification 'Elixir.KafkaEx.Protocol.JoinGroup':create_request(integer(),binary(),binary(),binary(),[binary()],integer()) -> binary() is a supertype of the success typing: 'Elixir.KafkaEx.Protocol.JoinGroup':create_request(integer(),binary(),binary(),binary(),[binary()],integer()) -> <<_:64,_:_*8>>` complaining about the return type `binary()` vs `<<_:64,_:_*8>>` (note the later is not a valid elixir typespec) which we should fix.

> Document the errors.

I need help with this, thanks!

eproxus:
> So there are a couple of concerns:
>
> * The kafka application (I mean application in erlang terms here) requires connection to kafka on start up by default (though you can bypass this via disable_default_worker: true config option), this error you cannot catch, since it happens before your code is loaded.

That's true. Although a connection error in that worker would crash the KafkaEx application right way, right? As I think it should.

> * If kafka really crashes hard you want the workers processing the messages to crash and not pretend all is well, I willing to be convinced about alternatives.

I totally see that point. Our dilemma comes when we don't want to restart these workers infinitely and without any backoff inbetween. We currently cannot do this with KafkaEx managing the workers for us. Even if the errors are reported to us correctly. The alternative proposed in this PR is to not crash them, but throw the errors out and let them run. I really agree that this is not so desirable as it might pollute their state and just make them a bit too perpetual.

> So this test underlines how kafka/zookeeper works, we need only connection to one broker to determine the state of the cluster so we go through each of the brokers supplied and attempt connection to them and only keep state of the ones we could successfully connect to, we also rebalance both async and sync based on the hosts provided by kafka/zookeeper this allows the user to add brokers to their cluster seamlessly and also remove a broker in a controlled manner without having to update their application configuration or restart their application.

If I then understand it correctly, a worker/server (the term seems to be used interchangeably) is a process inside KafkaEx that maintains connections to all brokers, and funnels all user requests (fetch/offset/etc.) to one of the brokers (according to what logic?). Is this correct?

bjhaid:
> I totally see that point. Our dilemma comes when we don't want to restart these workers infinitely and without any backoff inbetween

This can be configured via the worker proposed by @dantswain, this is also currently configurable via https://github.com/kafkaex/kafka_ex/blob/master/config/config.exs#L21-L24

> If I then understand it correctly, a worker/server (the term seems to be used interchangeably) is a process inside KafkaEx that maintains connections to all brokers, and funnels all user requests (fetch/offset/etc.) to one of the brokers (according to what logic?). Is this correct?

At initialization KafkaEx worker/server/gen_server establishes (attempts) an individual connection to all the brokers configured, and grabs kafka metadata from the broker it connects to first, this metadata includes information about hosts that the kafka cluster knows about. This metadata request is continuously requested asynchronously and synchronously to update the kafka brokers that the worker/server/gen_server knows about, this metadata also includes topic -> broker mapping which the worker/server/gen_server uses to determine which broker to dispatch requests to.

lib/kafka_ex/network/tcp.ex
bjhaid: Nice!

dantswain:
@eproxus 

> Our dilemma comes when we don't want to restart these workers infinitely and without any backoff inbetween. 

I'm curious what the use case is here.   We use KafkaEx in production at Simpli.fi for some fairly business-critical streaming and if an entire Kafka cluster is down we really do need the system to start screaming at us as loudly as possible... Our applications (in the broader sense, not the OTP sense) are almost universally under monit control and will indefinitely try to restart until everything is OK.  OTOH we have at least 3 nodes in all of our Kafka clusters and have never had an entire cluster be down.

One thing I'll add to the convo about metadata and nodes being down - It only really works if you have replication set up in your cluster.  If you need to fetch from a partition on a node that is down, KafkaEx will crash regardless of how many other nodes are in the cluster.

FYI - We have a channel (#kafkaex) on elixir-lang.slack.com if you'd like to chat more directly :)  Either @bjhaid or I are usually around during US daytime.

eproxus:
As per discussion on the Elixir slack channel:

Summary: the problem is that KafkaEx owns worker processes which currently cannot handle connection errors to brokers.

There are so far 3 identified ways to attack this problem:

1. Make KafkaEx provide the worker but not supervise it, leaving it up to the application to handle errors in the worker.
2. Make the worker more resilient to connection problems and report them out to the caller.
3. Expose a functional Kafka API that the workers internally use, but developers can use directly if they want to.

2 is somewhat complex, because errors needs to be reported out of the worker without crashing it, and the connections have to be reestablished by the worker.

3 is the most complex solution, and is problematic b/c the client state takes a lot of management and it would kind of break encapsulation there.

Therefore, for now, 1 seems to be the way to go.

bjhaid:
We can still make use of the good work done in this PR, we should be raising sane errors instead of the causing a pattern error failure of nil like we currently do, and catching the errors in places where we really don't need to crash

eproxus:
I think it is fairly straight forward to raise the errors outside, to the user, as in this proof of concept for fetch. However, are there any asynchronous cases where the worker talks to Kafka in the background? Should we just crash there?

When we don't crash, what to do with the broken socket? Reconnect or crash anyway?

bjhaid:
@eproxus I think we need 2 types of the network methods, one set with a`!` and another set without a `!`, the one with a `!` will raise and the other will not raise. Then we can use the one `!` in places where we want to crash and the one that does not raise in places where do not intend to crash.

eproxus:
@bjhaid Really good idea! I like it!

bjhaid:
@eproxus bump

eproxus:
Sorry, I've been really busy with other stuff at work at the moment. Cannot really give you an estimate of when I'm able to work further on this at the moment.

PR #128, opened 2016-06-21
Add KafkaEx.worker_spec/2

eproxus:
This adds an API function to return the KafkaEx worker child
specification for inclusion in custom supervisor structures.

**Note:** I did not update the documentation HTML since it was already outdated.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -46,6 +46,53 @@ defmodule KafkaEx do
   end
 
   @doc """
+  Returns a KafkaEx worker child specification.
+
+  This child specification can be used to host KafkaEx workers outside of the
+  KafkaEx application itself, for example in your own application supervision
+  tree. Use together with the application option `disable_default_worker` set
+  to `true` to make sure no processes are started by the KafkaEx
+  application.
+
+  It takes the same arguments as create_worker/2.
+
+  Returns {:error, error_description} on invalid arguments
+
+  ## Example
+
+  ```elixir
+  iex> KafkaEx.worker_spec(:pr) # where :pr is the name of the worker created
+  {:ok,
+   {KafkaEx.Server,
+    {KafkaEx.Server, :start_link,
+     [[uris: [{"localhost", 9092}, {"localhost", 9093}, {"localhost", 9094}],
+       consumer_group: "kafka_ex"], :pr]}, :permanent, 5000, :worker,
+    [KafkaEx.Server]}}
+
+  iex> KafkaEx.worker_spec(:pr, [uris: [{"localhost", 9092}], consumer_group: "foo", sync_timeout: 2000])
+  {:ok,
+   {KafkaEx.Server,
+    {KafkaEx.Server, :start_link,
+     [[uris: [{"localhost", 9092}], consumer_group: "foo", sync_timeout: 2000],
+      :pr]}, :permanent, 5000, :worker, [KafkaEx.Server]}}
+
+  iex> KafkaEx.worker_spec(:pr, consumer_group: nil)
+  {:error, :invalid_consumer_group}
+
+  ```
+
+  """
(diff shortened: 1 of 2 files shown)


test/kafka_ex_test.exs
```diff
@@ -1,6 +1,8 @@
 defmodule KafkaExTest do
   use ExUnit.Case, async: false
 
+  doctest KafkaEx, only: [worker_spec: 2]
```
bjhaid: Can we use real tests and not doctests, `server` will not be `KafkaEx.Server` going forward, it will be a more specific server to that works for the version of kafka you are running, so the test needs a dynamic value.

bjhaid:
We generate the HTML only when we release kafka_ex

test/kafka_ex_test.exs
eproxus: Fixed in the new commit.

lib/kafka_ex.ex
```diff
@@ -46,6 +46,53 @@ defmodule KafkaEx do
...
+
+  It takes the same arguments as create_worker/2.
+
+  Returns {:error, error_description} on invalid arguments
+
+  ## Example
+
+  ```elixir
+  iex> KafkaEx.worker_spec(:pr) # where :pr is the name of the worker created
+  {:ok,
```
dantswain: Maybe by way of example, just show this getting used in a supervisor?  All of the structure here doesn't really help now that this isn't a doctest.  Something like

```
{:ok, spec} = KafkaEx.worker_spec(:pr)
supervisor_opts = [strategy: :one_for_one, name: MySupervisor]
Supervisor.start_link(spec, supervisor_opts)
```

lib/kafka_ex.ex
```diff
@@ -46,6 +46,53 @@ defmodule KafkaEx do
...
+      :pr]}, :permanent, 5000, :worker, [KafkaEx.Server]}}
+
+  iex> KafkaEx.worker_spec(:pr, consumer_group: nil)
+  {:error, :invalid_consumer_group}
+
+  ```
+
+  """
+  @spec worker_spec(atom, KafkaEx.worker_init) :: {:ok, Supervisor.Spec.spec()} | {:eror, any}
+  def worker_spec(name, worker_init \\ []) do
```
dantswain: How would you feel about the name `build_supervisor_spec` or `worker_supervisor_spec` or some such?  "spec" is just kind of an overloaded term..

dantswain:
Pretty much exactly what I was thinking (aside from minor comments).  We could potentially add an `!` version that doesn't return the tuple with `:ok | :error`.

lib/kafka_ex.ex
eproxus: I'm not a big fan of verbs if the function returns a simple "item" (i.e. `get_value` or `return_result`). I think in this case, it is not a huge operation, therefore I'd like to avoid the heavy verb `build`. The other alternative should be `worker_child_spec` and here I'm of the opinion that it is a bit redundant. If you're after a child spec, you know you will put it in a supervisor, and a worker is normally not a supervisor, which implies it is just a normal child. The name "worker" in KafkaEx is unfortunate because it is the same as `Supervisor.Spec.worker`, so actually the proper canonical name should be `worker_worker_spec`. :smile:

lib/kafka_ex.ex
eproxus: Agree, I'll change to a better example.

eproxus:
A `worker_spec!/1` should raise an Elixir exception. How about `KafkaEx.Error.InvalidInitArgs`?

bjhaid:
@eproxus bump! any updates

eproxus:
Sorry for being unresponsive on this. I've recently switched projects (and job) so I haven't been using this for a while. Unfortunately, I currently won't have any time to work on KafkaEx going forwards.

PR #129, opened 2016-06-22
Make dialyzer happy.

bjhaid:
Some of the changes in this commits are annoying, but they are required
to make dialyzer happy, it defeats the purpose if we include a dialyzer
command in the README and yet we have a plenty dialyzer warnings.

@dantswain ^^  please review and merge if it's okay


bjhaid:
Good thing I PR'ed this, just remember why we didn't fix all the while, this will pass only on elixir >= 1.3.0

Issue #131, opened 2016-07-02
Add a moduledoc tag to document kafkaex.ex module


bjhaid:

dantswain:
Easiest way to do this is to use `@moduledoc File.read!(Path.expand("../README.md", __DIR__))`

bjhaid:
@dantswain thanks! can you make that change and commit it?

bjhaid:
closed by #132

PR #132, opened 2016-07-03
Minor doc updates

dantswain:
As per #131 

* Set KafkaEx moduledoc to main README.md
* Set main ExDoc page to KakfaEx

@bjhaid Should I also generate and commit the docs?


bjhaid:
@dantswain thanks! You probably didn't need to PR this, I usually delay generating docs till when I am creating a release, but given how this impacts the docs, I'll say go for it and check how it looks locally then you can :shipit:

PR #135, opened 2016-07-12
KafkaEx.produce returns {:ok, offset} when required_acks != 0

### LINKED ISSUE #133: Produce request with required_acks <> 0 returns raw result
Contrary to what the @doc on `KafkaEx:produce` specifies, using a non-zero value for `required_acks` does not return a simple error/ok indication; it returns the (internal) ProduceResult instead. Some logic should be added to parse that down to a simple indication.

divtxt:
Fixes #133 

Also fixes typos where required_acks parameter was sometimes spelt as require_acks.

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -288,6 +288,8 @@ defmodule KafkaEx.Server do
               case response do
                 [%KafkaEx.Protocol.Produce.Response{partitions: [%{error_code: :no_error, offset: offset, partition: _}], topic: topic}] when offset != nil ->
                   {:ok, offset}
+                _ ->
+                  {:error, response}
               end
           end
         end
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -281,9 +281,14 @@ defmodule KafkaEx.Server do
             :leader_not_available
           broker -> case produce_request.required_acks do
             0 ->  NetworkClient.send_async_request(broker, produce_request_data)
-            _ -> broker
-            |> NetworkClient.send_sync_request(produce_request_data, state.sync_timeout)
-            |> Produce.parse_response
+            _ ->
+              response = broker
+               |> NetworkClient.send_sync_request(produce_request_data, state.sync_timeout)
+               |> Produce.parse_response
+              case response do
+                [%KafkaEx.Protocol.Produce.Response{partitions: [%{error_code: :no_error, offset: offset, partition: _}], topic: topic}] when offset != nil ->
+                  {:ok, offset}
+              end
           end
         end
         state = %{state | correlation_id: corr_id + 1}
--- a/test/integration/integration_test.exs
+++ b/test/integration/integration_test.exs
@@ -104,8 +104,7 @@ defmodule KafkaEx.Integration.Test do
   end
 
   test "produce/4 with ack required returns an ack" do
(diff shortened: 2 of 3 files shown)


bjhaid:
thanks!, I will review sometimes soon

lib/kafka_ex/server.ex
```diff
@@ -281,9 +281,16 @@ defmodule KafkaEx.Server do
...
+            _ ->
+              response = broker
+               |> NetworkClient.send_sync_request(produce_request_data, state.sync_timeout)
+               |> Produce.parse_response
+              case response do
+                [%KafkaEx.Protocol.Produce.Response{partitions: [%{error_code: :no_error, offset: offset, partition: _}], topic: topic}] when offset != nil ->
+                  {:ok, offset}
+                [%KafkaEx.Protocol.Produce.Response{partitions: [%{error_code: 0, offset: offset, partition: _}], topic: topic}] when offset != nil ->
+                  {:ok, offset}
+              end
```
cdegroot: I think that you really want a wildcard clause here to return something like `{:error, response}`. \

Are both success code paths possible here?

divtxt: Added wildcard clause returning `{:error, response}`.

Yes both success code paths are possible. I've seen `:no_error` in tests, and we see `0` in our service.

lib/kafka_ex/server.ex
```diff
@@ -281,9 +281,16 @@ defmodule KafkaEx.Server do
...
             0 ->  NetworkClient.send_async_request(broker, produce_request_data)
-            _ -> broker
-            |> NetworkClient.send_sync_request(produce_request_data, state.sync_timeout)
-            |> Produce.parse_response
+            _ ->
+              response = broker
+               |> NetworkClient.send_sync_request(produce_request_data, state.sync_timeout)
+               |> Produce.parse_response
+              case response do
+                [%KafkaEx.Protocol.Produce.Response{partitions: [%{error_code: error_code, offset: offset, partition: _}], topic: topic}] when (error_code == 0 or error_code == :no_error) and offset != nil ->
```
bjhaid: `error_code` can only be `:no_error` see: https://github.com/kafkaex/kafka_ex/commit/37f6ddd9760d615b48e84214f0b86b9bc7d5e1b6

bjhaid:
thanks for the work, other than my comment, LGTM, I will merge as soon as you update `error_code` ❤️

lib/kafka_ex/server.ex
divtxt: Got it! Looks like 0.5.0 release was before this commit which may be why we're getting `0`.

divtxt:
Fixed

test/integration/integration_test.exs
```diff
@@ -115,10 +114,7 @@ defmodule KafkaEx.Integration.Test do
   end
 
   test "produce with ack required returns an ack" do
-    produce_response = KafkaEx.produce(%Proto.Produce.Request{topic: "food", partition: 0, required_acks: 1, messages: [%Proto.Produce.Message{value: "hey"}]}) |> hd
-    offset = produce_response.partitions |> hd |> Map.get(:offset)
-
-    refute offset == nil
```
bjhaid: this assertion has been deleted can you please re-add it, thanks

divtxt: My bad. Added back

PR #136, opened 2016-08-04
Call kafka_server_* functions locally

pguillory:
After calling `use KafkaEx.Server`, the handle_call/info callbacks are inserted
into the current module. Thus they can call kafka_server_* functions locally,
rather than as server_impl.kafka_server_*. This saves the trouble of storing
server_impl in the GenServer state.


pguillory:
Ah, failed to run the integration test before submitting. This doesn't work.

pguillory:
Re-opening with fixes to the integration and consumer_group tests. All tests now pass. `dialyze` shows 70 warnings, but that appears already to have been the case on master.

PR #146, opened 2016-10-07
Add SSL support

comogo:
Hey guys,

Instead we handle socket directly in many places I created the `KafkaEx.Socket` module, so we'll have just one place to handle it.
This makes easier to change the code and help me to add support to SSL connection.

The erlang interface for the client side SSL and non SSL sockets are pretty much the same so
this helps a lot the SSL implementation. =D


bjhaid:
thanks alot, I like this implementation @dantswain @cdegroot what do you folks think?

bjhaid:
@comogo can you include an integration test that depicts using the ssl socket with kafka 0.10.0?

comogo:
I can try, I just need to learn how to set up the ssl certificate on kafka.

bjhaid:
http://docs.confluent.io/2.0.0/kafka/ssl.html can be useful

cdegroot:
I'm about to board a flight so cannot check this PR, alas. Skimming it I say it looks good and indeed, an integration test would be nice (but hairy. Java and TLS and certs... maybe check in a snake oil cert to make it easy to setup on the Kafka side?)

dantswain:
An integration test would be good, I agree.

I'm a little concerned about the options handling.  I think the SSL options should probably be configurable on a per-worker basis because some people may be connecting to multiple broker clusters with different configurations.  If supporting that is a burden at this point, we could let this PR go through and follow it up with another one that deals with the configuration, but it should be done sooner than later.

comogo:
I agree with @dantswain  so I added one more commit to provide SLL options per-worker.
So, what do you think about that guys?

comogo:
I have no enough time to configure a kafka with SSL for the integration test yet.

bjhaid:
@comogo thanks I merged this, I will find sometime to add the integration specs

PR #151, opened 2016-11-11
Use earliest available offset on partition

dplummer:
If you use offset 0 for a topic that's been around a while, it might not be available. At least on Kafka 0.8.2 you get an error `:offset_out_of_range` and no messages are returns.

This PR uses `earliest_offset` to get the first that's available.

I don't know if tests pass, as I am still working on getting `snappy` to compile on my machine. Until I do, I'd like input on if you think this change is a good idea.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -339,7 +339,8 @@ Optional arguments(KeywordList)
           |> OffsetFetchResponse.last_offset
 
         if last_offset <= 0 do
-          0
+          earliest_offset(topic, partition, worker_name)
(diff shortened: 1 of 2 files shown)


bjhaid:
your PR will not prevent the case of using `offset: 0` for a topic that has been around, since you are modifying the case where no offset is supplied, also I don't think we should be trying to generate the offset for the user when they explicitly pass in the offset it will break compatibility with kafka < 0.8.2

dplummer:
I got tests running locally by using a vagrant box. They all pass for me, I'm unable to reproduce the error travis is seeing.

PR #154, opened 2016-11-27
Use range strategy for join group request

mlewis033:
- set strategy_name to "range"
- added a length header for the ProtocolMetadata blob

Tested with kafka_2.11-0.9.0.1 kafka_console_consumers sharing the same consumer_group using manual join_group and sync_group requests in non-group-leader mode.

### DIFF
--- a/lib/kafka_ex/protocol/join_group.ex
+++ b/lib/kafka_ex/protocol/join_group.ex
@@ -5,7 +5,7 @@ defmodule KafkaEx.Protocol.JoinGroup do
   Implementation of the Kafka JoinGroup request and response APIs
   """
   @protocol_type "consumer"
-  @strategy_name "assign"
+  @strategy_name "range"
   @metadata_version 0
 
   defmodule Request do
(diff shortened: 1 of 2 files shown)


bjhaid:
@mlewis033 thanks for this PR, can you instead of changing the strategy, make it such that the user can pass it in, and default it to something when not supplied and also ensure we are validating the input. @cdegroot can also comment since he did most of this work.

cdegroot:
@mlewis033 what is the purpose of this change? "range" and "roundrobin" are two valid strategies only in the context of the Java client library; unless you want to have mixed Elixir/Java consumers in the same group, you can put here whatever you want and - unless I'm overlooking something - it's not going to make a difference.

mlewis033:
@cdegroot this change is the result of examining behavior with kakfa_console_consumer consumer_group peers hence the bias towards the default [RangeAssignor](https://github.com/apache/kafka/blob/0.9.0/clients/src/main/java/org/apache/kafka/clients/consumer/RangeAssignor.java) strategy of that client.  I'd assumed that it was the 'reference' strategy.

When specifying that strategy in this mixed-platform environment, the current KafkaEx ProtocolMetadata encoding crashes the java client when it tries to SyncGroup.  The addition of a length-header fixes that issue.

cdegroot:
Given that partition assignments are client-side, the wording change from "assign" to "range" doesn't do a thing. I don't like letting things through because magic - if you need mixed Java/Elixir consumers, there's more work to do (the Elixir consumer might become the group leader and thus has to make sure that it implements the whole Java assignment protocol). 

+1 on the length header patch - the docs do recommend keeping private protocol data as much compatible with the Java client as possible, because some utilities interpret the data.

dantswain:
@mlewis033 @cdegroot Can we get some resolution on this?  Especially if the header patch may help some folks.

cdegroot:
I think we can close this one - another patch should have solved this (#162 or #175, IIRC)

PR #157, opened 2017-01-05
Reduce logging level when connecting to broker

davidrusu:
Hi, Thank you for the great library!

We are trying to reduce the amount of logging our application is producing and this log in particular is generating on the order of 100k logs per hour.

```"Establishing connection to broker #{metadata_broker.node_id}: #{inspect metadata_broker.host} on port #{inspect metadata_broker.port}"```

Do you mind reducing the logging level to :debug so that we may compile it out when we run in production?

Thanks,

David

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -383,7 +383,7 @@ defmodule KafkaEx.Server do
         case length(brokers_to_keep) do
           0 -> brokers_to_remove
           _ -> Enum.each(brokers_to_remove, fn(broker) ->
-            Logger.log(:info, "Closing connection to broker #{broker.node_id}: #{inspect broker.host} on port #{inspect broker.port}")
+            Logger.log(:debug, "Closing connection to broker #{broker.node_id}: #{inspect broker.host} on port #{inspect broker.port}")
             NetworkClient.close_socket(broker.socket)
           end)
             brokers_to_keep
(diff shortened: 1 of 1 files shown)


dantswain:
I'm not opposed to this.  Would it be even better to make it configurable?  @bjhaid ?

bjhaid:
@dantswain 👍  to both ideas are you taking making it configurable?

davidrusu:
I don't see much value (at least for us) in making this configurable, this would be enough for us.

dantswain:
@davidrusu Could you please also change the `Logger.log(:info, ..)` on line 386 to be `:debug`?

@bjhaid I think let's merge this (after the requested change) and put some thought into the configurability (which logs should be configurable, etc) as a future change.

davidrusu:
@dantswain Done! Let me know if you need anything else.

dantswain:
:tada: Thanks, @davidrusu!

davidrusu:
@dantswain Just wanted to make you aware of this flag in Logger `:compile_time_purge_level`. You can see the docs here https://hexdocs.pm/logger/Logger.html

Instead of cluttering the codebase with conditional logging, consider suggesting to users to use this flag in their application?

davidrusu:
@dantswain @bjhaid I just realized I didn't read the documentation carefully enough!

This `:compile_time_purge_level` flag only applies to Logger.(debug | info | ...) calls, not Logger.log calls.

Would you mind a PR that changes the Logger calls to use these macro calls instead of `Logger.log`?

dantswain:
@davidrusu I would prefer not making that change right now unless you have some evidence that the shunted `Logger.log` calls are impacting performance.  I do think our logging could use some review, so I created #163 to channel efforts and scope.

Issue #159, opened 2017-01-06
Not handling MemberMetadata when parsing JoinGroupResponse


dcuddeback:
In [`JoinGroupResponse`](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-JoinGroupResponse), each element of the `Members` array contains two fields. The implementation right now only parses the `MemberId` field. Even if `MemberMetadata` is empty, it has at least a 4-byte length field, which is being skipped while parsing the response.

Issue #158, opened 2017-01-06
Missing metadata length in serialized JoinGroupRequest


dcuddeback:
In [`JoinGroupRequest`](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-JoinGroupRequest), `ProtocolMetadata` is defined as `bytes`. `bytes` is defined to have a 32-bit length field preceding it: https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-TheProtocol. This length field is missing from the serialized request.

PR #160, opened 2017-01-06
Fix JoinGroup Serialization

dcuddeback:
This PR fixes the serialization errors described in #158 and #159.

### DIFF
--- a/lib/kafka_ex/protocol/join_group.ex
+++ b/lib/kafka_ex/protocol/join_group.ex
@@ -63,8 +63,10 @@ defmodule KafkaEx.Protocol.JoinGroup do
               leader_id: leader, member_id: member_id, members: members}
   end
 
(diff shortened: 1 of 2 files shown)


bjhaid:
@cdegroot please do review thanks!

dantswain:
Did the spec change?  That's my main concern.

cdegroot:
@dantswain if you meant the Kafka spec, no. That spec is actually very well-versioned. I typed in most of the serialization code in one go, with in the back of my mind that it doesn't really matter too much whether it's correct _unless_ we want to aim for 100% compatibility with the Java client. So @dcuddeback is just fixing my mistakes here, most likely ;-).

cdegroot:
PR is cleanup and adding a missing byte size field, which is in line with the spec for `ProtocolMetadata` (specified as `bytes` so it needs a size). :shipit:

PR #172, opened 2017-01-12
Add coveralls test coverage reporting

dantswain:
* Add excoveralls to the project
* Some build matrix updates to get this to work (excoveralls doesn't work with R17)
* Move the test running to a script which makes it easier to deal with env vars
* Simple formatting for mix.exs


dantswain:
Heh.  Let me try to figure out a way for us to only get one report.

PR #176, opened 2017-01-13
Cleanup code duplicated between Server0P9P0 and Server0P8P2.

bjhaid:


lib/kafka_ex/server_0_p_9_p_0.ex
```diff
@@ -29,6 +25,15 @@ defmodule KafkaEx.Server0P9P0 do
     GenServer.start_link(__MODULE__, [args, name], [name: name])
   end
 
+  defdelegate kafka_server_consumer_group(state), to: Server0P8P2
```
dantswain: May be worth adding a comment here to the effect that the implementation for these is the same in 0.9.0 as it is in 0.8.2?

bjhaid:
@dantswain does the comment look good?

PR #179, opened 2017-01-14
Validate ssl options, relax use_ssl false condns

### LINKED ISSUE #178: Setting `use_ssl` to false without changing `ssl_options` causes a crash
From this section of the config:

```
  # This is the flag that enables use of ssl
  use_ssl: true,
  # see SSL OPTION DESCRIPTIONS - CLIENT SIDE at http://erlang.org/doc/man/ssl.html
  # for supported options
  ssl_options: [
    cacertfile: System.cwd <> "/ssl/ca-cert",
    certfile: System.cwd <> "/ssl/cert.pem",
    keyfile: System.cwd <> "/ssl/key.pem",
  ],
```

I would expect that setting `use_ssl` to false will simply make `ssl_options` irrelevant.  Instead it causes the application to fail to start with a 

```
** (Mix) Could not start application kafka_ex: KafkaEx.start(:normal, []) returned an error: :badarg
```

which is pretty unhelpful.

I propose:

1. If `use_ssl` is set to false, the application should ignore `ssl_options`.
2. If anything in `ssl_options` is configured incorrectly, the application should raise a meaningful error to guide the user in determining what is wrong.

dantswain:
Closes #178

I'm not an ssl expert so wasn't sure what else to do for validation.  This checks for `:certfile` `:cacertfile` and `:keyfile` being set and pointing to readable files.  If anyone has input on that it'd be appreciated!

### DIFF
--- a/lib/kafka_ex/config.ex
+++ b/lib/kafka_ex/config.ex
@@ -1,6 +1,15 @@
 defmodule KafkaEx.Config do
   @moduledoc false
 
+  require Logger
+
+  def use_ssl, do: Application.get_env(:kafka_ex, :use_ssl)
+  # use this function to get the ssl options - it verifies the options and
+  #   either emits warnings or raises errors as appropriate on misconfiguration
+  def ssl_options do
+    ssl_options(use_ssl(), Application.get_env(:kafka_ex, :ssl_options))
+  end
+
   def default_worker do
     :kafka_ex
   end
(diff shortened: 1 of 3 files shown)


dantswain:
@comogo If you have a few minutes, would you take a look at this and let me know if I've done anything incorrect / bad with respect to the ssl options?  Thanks!

dantswain:
@comogo I'm going to merge this because I need it to move forward on something else, but I'd still appreciate any feedback you have.

bjhaid:
This change will prevent configuring SSL on a per worker basis

bjhaid:
Sorry I noticed that late

dantswain:
@bjhaid How's that?  The mechanism is still the same, it just goes through a validation function first.

dantswain:
Oh nevermind, I see it now.  OK, I will fix that ASAP.

PR #180, opened 2017-01-14
Fix ssl configuration per worker

dantswain:
Fixes a problem I accidentally introduced in #179 

Also:
* Fix `ssl_options` type
* Remove validation of the presence of the certfiles in ssl_options - It looks like there are other valid combinations of options here.  We still validate that the files are readable if they are present in the options and use_ssl is set to true.


bjhaid:
LGTM :shipit:

dantswain:
I'd like to add a test to get the coverage back up before merging this.

PR #181, opened 2017-01-16
POC testing with Docker

dantswain:
**DO NOT MERGE**

This PR is for feedback purposes.  After addressing feedback, I want to do a squashed commit and re-submit.

This adds configuration necessary to launch a containerized SSL-enabled Kafka 0.9 cluster for testing and uses that for testing on travis.

Some open questions:

* With this in place, can we remove the existing scripts for installing/setting up Kafka?  I understand wanting to support those who don't have docker, but I kind of think the standardization is worth it.  Especially now that the docker for mac experience has improved significantly.

* I would like to add tests that exercise the non-ssl code paths.  This might be achievable by adding another set of listeners to the cluster.

* We have talked about testing multiple versions of Kafka, which Docker should help with a lot.  I have put some thought into how to do that, but this PR doesn't really reflect that.  I think the only sane way to do that will be to launch multiple clusters simultaneously and test using separate sets of connections so that we don't have to add a ton of code to the test suite to handle starting different clusters.  I don't think it would be good to do that using the travis build matrix (e.g., separate builds) because a) it would exponentially grow the size of the build matrix and b) we wouldn't get a single coherent coverage report.

If the approach in this PR is agreeable, I'll take this as a starting point to address testing non-SSL and different Kafka versions.


.travis.yml
```diff
@@ -20,16 +20,17 @@ matrix:
     - elixir: 1.4.0
       otp_release: 19.2
       env: COVERALLS=true
+dist: trusty
 sudo: required
+services:
+  - docker
 env:
   - COVERALLS=false
 cache:
```
dantswain: I think we could get rid of this now?

scripts/docker_up.sh
```diff
@@ -0,0 +1,40 @@
...
+  sed -i.bak "s/@broker_id@/${i}/g" ${target}
+  sed -i.bak "s/@port@/${port}/g" ${target}
+  # the @pwd become root directory so we get /ssl
+  sed -i.bak "s|@pwd@||g" ${target}
+  # point at zookeeper in docker
+  sed -i.bak "s/localhost:2181/zookeeper:2181/" ${target}
+  # delete the existing listeners line
+  sed -i.bak "/^listeners=/d" ${target}
+  # add an advertised.listeners and listeners together at the end
+  echo "advertised.listeners=SSL://${DOCKER_IP}:${port}" >> ${target}
```
bjhaid: this can be killed based on my previous comment and we can also kill `IP_IFACE` and `DOCKER_IP` along with this

lib/kafka_ex/server.ex
```diff
@@ -391,7 +391,7 @@ defmodule KafkaEx.Server do
                        Metadata.parse_response(data)
                    end
 
-                   case Enum.find(response.topic_metadatas, &(&1.error_code == :leader_not_available)) do
+        case Enum.find(response.topic_metadatas, &(&1.error_code == :leader_not_available)) do
```
bjhaid: is the ident change necessary?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      - ./scripts/docker-start-kafka.sh:/usr/bin/start-kafka.sh
+      - ./ssl:/ssl
+  kafka2:
+    image: wurstmeister/kafka:0.9.0.1
+    ports:
+      - "9093:9093"
+    depends_on:
+      - zookeeper
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
bjhaid: this is not necessary

scripts/docker_up.sh
```diff
@@ -0,0 +1,40 @@
...
+  cp ./server.properties ${target}
+  # configure broker and port
+  sed -i.bak "s/@broker_id@/${i}/g" ${target}
+  sed -i.bak "s/@port@/${port}/g" ${target}
+  # the @pwd become root directory so we get /ssl
+  sed -i.bak "s|@pwd@||g" ${target}
+  # point at zookeeper in docker
+  sed -i.bak "s/localhost:2181/zookeeper:2181/" ${target}
+  # delete the existing listeners line
+  sed -i.bak "/^listeners=/d" ${target}
```
bjhaid: we should use listeners with `0.0.0.0` instead of `advertised.listeners` as below which is probably more brittle

scripts/docker-start-kafka.sh
```diff
@@ -0,0 +1,57 @@
...
+if [[ -n "$KAFKA_HEAP_OPTS" ]]; then
+    sed -r -i "s/(export KAFKA_HEAP_OPTS)=\"(.*)\"/\1=\"$KAFKA_HEAP_OPTS\"/g" $KAFKA_HOME/bin/kafka-server-start.sh
+    unset KAFKA_HEAP_OPTS
+fi
+
+# any env vars starting with KAFKA will be set in server.properties
+#   e.g., KAFKA_LOG_DIRS =>log.dirs
+# if the var is already in server.properties, it will be replaced
+# if it is not already in server.properties, it will be added to the end
+for VAR in `env`
```
bjhaid: the env var substitution is not necessary since we are not setting them anywhere, same with the `KAFKA_HEAP_OPTS` from the previous `if` condition

.travis.yml
bjhaid: maybe?

bjhaid:
>>> With this in place, can we remove the existing scripts for installing/setting up Kafka? I understand wanting to support those who don't have docker, but I kind of think the standardization is worth it. Especially now that the docker for mac experience has improved significantly.

I believe this should become the defacto and we can kill the other scripts, since we can easily setup more kafka versions that we can test against with this

I would like to add tests that exercise the non-ssl code paths. This might be achievable by adding another set of listeners to the cluster.

👍  but in a different PR

We have talked about testing multiple versions of Kafka, which Docker should help with a lot. I have put some thought into how to do that, but this PR doesn't really reflect that. I think the only sane way to do that will be to launch multiple clusters simultaneously and test using separate sets of connections so that we don't have to add a ton of code to the test suite to handle starting different clusters. I don't think it would be good to do that using the travis build matrix (e.g., separate builds) because a) it would exponentially grow the size of the build matrix and b) we wouldn't get a single coherent coverage report.

I agree with this approach, though we need to mess with the test code to support this but it shouldn't be that complicated

lib/kafka_ex/server.ex
dantswain: No but it was incorrect before.  I just happened to be in this file debugging.  I can remove it because it's irrelevant but it should be fixed.

bjhaid: cool I think this should go in a separate commit/PR

scripts/docker-start-kafka.sh
dantswain: I argued with myself about this a bit and had intentionally decided to leave it in (note I stripped a bunch of other stuff).  Some of these settings can be handy to be able to modify, but I guess the approach of copying in the server.properties file this isn't as big of a deal.  I can remove it.

scripts/docker_up.sh
dantswain: Combined response with the comment on IP_IFACE/DOCKER_IP.

It doesn't work without this.  It took me a few hours to figure this out :(  The `listeners` setting has to specify binding to 0.0.0.0 to get SSL to work, but we have to advertise DOCKER_IP so that the nodes are able to talk to one another _and_ so that the outside world is able to communicate with the cluster.

dantswain:
@bjhaid I think I addressed your feedback.  Should I remove the old install/setup scripts as part of this change?  (again - I'm going to squash all this into a single commit and open a separate PR with that, but I want to make sure all of the changes are clean first)

bjhaid:
There's a `squash and merge` button on github UI that can help you with squashing :) 
image

PR #188, opened 2017-01-27
Add config.exs to the package, version bump

dantswain:
This is needed because the config module reads config.exs to generate
documentation.


dantswain:
Thanks, good catch!

Issue #191, opened 2017-02-12
stream/3 hangs


kevinbader:
Hi all,

I'm not sure whether I don't understand how stream/3 is supposed to work, or it's buggy:

```
defmodule KafkaTest do
  @moduledoc false
  @topic "elixirtest"
  @partition 0

  # this works
  def uses_stream_immediately_takes_two do
    KafkaEx.stream(@topic, @partition, offset: 0)
    |> Stream.map(& "the value: " <> &1.value)
    |> Enum.take(2)
  end

  # using to_list instead of take causes the method to never return:
  def uses_stream_immediately_takes_all do
    KafkaEx.stream(@topic, @partition, offset: 0)
    |> Stream.map(& "the value: " <> &1.value)
    |> Enum.to_list
  end

  # not using the stream immediately causes the method to never return, too:
  def waits_before_using_the_stream_takes_two do
    theStream = KafkaEx.stream(@topic, @partition, offset: 0)
    :timer.sleep(1000)  # waiting a bit..
    theStream
    |> Stream.map(& "the value: " <> &1.value)
    |> Enum.take(2)
  end
end
```

dantswain:
I think this is a consequence of how `stream/3` currently works.  @bjhaid may have more info, but we have #185 to manage a rewrite.

bjhaid:
It's due to the implementation of stream in kafka_ex, it uses GenEvent for streaming which we intend to replace in the issue @dantswain linked and I believe it's related to this: https://github.com/elixir-lang/elixir/issues/3138.

I have some WIP branch for #185 that I'll invest more time in this week to complete

dantswain:
Hi @kevinbader The stream rewrite is complete.  If you have a chance could you check out the latest master and see if it still exhibits this behavior?

dantswain:
I'm going to close this unless we get a confirmation that the problem persists after the stream rewrite.

Issue #185, opened 2017-01-21
Reimplement streaming without GenEvent


dantswain:
Per discussion with @bjhaid.  I think `Stream.resource/3` would do the trick https://hexdocs.pm/elixir/Stream.html#resource/3

This should address #145 and #141

Issue #145, opened 2016-10-04
auto_commit not working in KafkaEx.stream/2


gjaldon:
I'm using Kakfa 0.10.0 and have the following code:

``` elixir
for message <- KafkaEx.stream(@topic, partition) do
  process_message(pool, message)
end
```

The issue is that offsets are not being committed. I tried disabling the `auto_commit` option and then manually creating a worker and then manually committing the offset after the task but it still doesn't work. I based the offset_commit code on your test for the offset_commit function.

Also, `process_message/2` above takes about half a minute to finish. In case, that matters. 

What could be causing this problem?

bjhaid:
@gjaldon sorry for the late response, have you tried `KafkaEx.fetch` do you have similar problems, I will attempt to reproduce sometimes today

gjaldon:
@bjhaid no problem at all. I ended up getting the last offset from `KafkaEx.fetch` and passed the offset there to `KafkaEx.stream` as option so that it would start from that offset as a workaround.

spscream:
try to use a named consumer group as a workaround, worked for us

dantswain:
@gjaldon Can you provide a more complete reproduction case? I'd like to figure out what's going on here.  Thanks!

dantswain:
@bjhaid were you ever able to reproduce this?

bjhaid:
I didn't try to :man_facepalming:

Issue #141, opened 2016-09-08
Memory leak when using KafkaEx.stream


dplummer:
I mentioned on slack:

> I’ve written an app (https://github.com/avvo/kafkamon) that consumes kafka using KafkaEx. I seem to have a memory leak using `KafkaEx.stream` (https://github.com/avvo/kafkamon/blob/master/lib/reader/event_queue/consumer.ex#L25). Has anyone used that api and seen a similar issue?

And @dantswain suggested I create an issue after reproducing it with a minimal app. I built [KafkaConsumerTest](https://github.com/dplummer/kakfa_consumer_test) and was able to reproduce the behavior I saw reliably. The README in the project outlines how to run the consumer and producer.

I reproduced this with:

```
Erlang/OTP 19 [erts-8.0.2] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.3.2
```

And both v0.5.0 and github master of KafkaEx. My example producer creates a random binary, but the same behavior is observed for a string.
## Observed behavior on my MacBook Pro

Ran the producer twice and captured the memory and process graphs from the observer. Memory did not decrease after 5 minutes of no activity, except in Elixir.KafkaEx.Server, which did decrease to ~143k (from 264k).

![memory_usage](https://cloud.githubusercontent.com/assets/98884/18335974/6a962d28-7538-11e6-8f83-aa6601aaa15b.png)
![process_list](https://cloud.githubusercontent.com/assets/98884/18335976/6f2dfa78-7538-11e6-8589-2bf83978743f.png)

dplummer:
I tried using `KafkaEx.fetch` and did not observer a memory leak:

```elixir
  def init(:ok) do
    GenServer.cast self, :fetch
    {:ok, latest_offset(@topic)}
  end

  def handle_cast(:fetch, offset) do
    Logger.info "Fetch of #{@topic}##{offset}"

    [%{partitions: [%{message_set: messages}]}] = KafkaEx.fetch(@topic, 0, offset: offset)

    new_offset = messages
    |> Enum.reduce(0, fn message, max ->
      current_offset = process_message(message)
      cond do
        current_offset > max -> current_offset
        true -> max
      end
    end)

    Logger.info "Done with our fetch, new_offset is #{new_offset}"
    Process.sleep 1000
    GenServer.cast self, :fetch
    {:noreply, new_offset}
  end
```

This is after processing 250k messages:
[...]

PR #193, opened 2017-02-18
Stream

bjhaid:
This replaces the existing GenEvent based stream implementation with a purely lazy stream implementation, fixing issues described in:

#191
#185
#145 
#141


ook:
Hi, I switched one of my project on that branch, and the new `steam/3` implementation behave weirdly when reaching the latest offset: it appears `stream/3` stop streaming and return, which is not the expected behavior.

bjhaid:
@ook thanks for exercising the code, this behavior is documented here:

https://github.com/kafkaex/kafka_ex/blob/stream/lib/kafka_ex.ex#L260

We have a choice of allowing the stream to be finite or infinite, I chose the former, maybe we should allow the stream function take a flag to determine if it's a finite stream or an infinite one and act accordingly, then you can either have a stream that never returns like you want or one that returns once it's done consuming the available offsets like was mentioned in some other issue.

ook:
Both aspect are interesting. May I suggest a big warning in the `README.md` if you keep this behavior as default, since it breaks with the previous implementation?
I like the new implementation which permit to process tidying / update operations, eventually a sleep, when we reach the "end" of the stream.

lib/kafka_ex/stream.ex
```diff
@@ -0,0 +1,44 @@
+defmodule KafkaEx.Stream do
+  @moduledoc false
+  alias KafkaEx.Protocol.OffsetCommit.Request, as: OffsetCommitRequest
+  alias KafkaEx.Protocol.Fetch.Request, as: FetchRequest
+
+  defstruct worker_name: nil, fetch_request: %FetchRequest{}, consumer_group: nil
+
+  defimpl Enumerable do
+    def reduce(data, acc, fun) do
+      next_fun = fn offset ->
```
dantswain: Can we make this an actual function and/or refactor it into a composition of smaller functions?

lib/kafka_ex/stream.ex
```diff
@@ -0,0 +1,44 @@
...
+              consumer_group: data.consumer_group,
+              topic: data.fetch_request.topic,
+              partition: data.fetch_request.partition,
+              offset: offset, metadata: ""
+            }
+          })
+        end
+        response = data.worker_name
+        |> GenServer.call({:fetch, %{data.fetch_request| offset: offset}})
+        |> hd |> Map.get(:partitions) |> hd
```
dantswain: I'd prefer not to use `hd` so much - it's a bit opaque.  We can either pattern match `[element] = thing` or add helper functions to the appropriate modules, like `FetchResponse.first_response`.  Also I think we should try to stick to one pipe per line (I think credo would tell us this as well).

ook:
(FYI, changed into a recursive call to stream: works like a charm)

dantswain:
@bjhaid Any plans to revisit this?

bjhaid:
Will do sometime this week, I have not had enough time that I can dedicate to finishing this, but will try to find time to do this week

dantswain:
OK no worries, I just thought I'd ping you.   Let me know if there's anything I can do to help.

dantswain:
@bjhaid Monthly ping on this ;)

bjhaid:
@dantswain I will try to finish this today, and get it behind me

lib/kafka_ex.ex
```diff
@@ -253,18 +254,13 @@ Optional arguments(KeywordList)
...
   @doc """
-  Returns a stream that consumes fetched messages.
-  This puts the specified worker in streaming mode and blocks the worker indefinitely.
-  The handler is a normal GenEvent handler so you can supply a custom handler, otherwise a default handler is used.
-
-  This function should be used with care as the queue is unbounded and can cause OOM.
-
+  Returns a stream that consumes fetched messages, the stream will halt once the max_bytes number of messages is reached, if you want to halt the stream early supply a small max_bytes, for the inverse supply a large max_bytes.
   Optional arguments(KeywordList)
+  - stream_mode: the mode the stream will be in, `:infinite` for an infinite stream and `:finite` for a finite stream, a finite stream will return once there's no data left to consume from the partition while an infinite stream will block till there's data to consume from the stream.
```
dantswain: I think it would be good if this mirrors the command line tools and uses something like `:no_wait_at_logend` https://cwiki.apache.org/confluence/display/KAFKA/System+Tools#SystemTools-SimpleConsumerShell

dantswain:
@bjhaid I've been playing with this locally and I realized that repeated `Enum.take` calls return the same messages:

```
iex> Enum.take(stream, 3)
[%KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3853573778, key: "",
  offset: 3, value: "WOO"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3853573778, key: "",
  offset: 4, value: "WOO"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3853573778, key: "",
  offset: 5, value: "WOO"}]

# note same offsets
iex> Enum.take(stream, 3)
[%KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3853573778, key: "",
  offset: 3, value: "WOO"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3853573778, key: "",
  offset: 4, value: "WOO"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3853573778, key: "",
  offset: 5, value: "WOO"}]
```

I've thought quite a bit about this and I _think_ that makes sense.  The `stream` object has not changed between calls.  OTOH I can see someone having an expectation that this would work, and I believe it's possible to _make_ it work with some changes.

I think what makes most sense is to provide a function in the Stream module that can update the stream struct's max offset so that it _can_ be used recursively - something like

```
def consume_from_stream(stream, handler) do
  messages = Enum.take(stream, 3) # or whatever
  handler.(messages)
  stream_out = KafkaEx.Stream.update_offset(stream, messages)
  consume_from_stream(stream_out, handler)
end
```

Does that make sense to you?  Also @joshuawscott do you have any thoughts?

bjhaid:
It is either the offset does not get committed until you have consumed the entire bytes returned in the last request made to kafka or `auto_commit` is disabled. If you manually commit the offset `Enum.take` will start returning something else. This is weird behavior but kafka streams don't map directly to a list or an elixir stream so maybe this should be documented.

dantswain:
Even with auto commit set to true, the fetch offset gets set when the stream object is created, then the fetch calls use that offset - so it ends up using the same offset every time.  We would have to remove that value from the initialization to get the next available offset each time.

bjhaid:
Then it's:

> the offset does not get committed until you have consumed the entire bytes returned in the last request made to kafka 

I still think it should be documented that the offset is not committed till the entire bytes window received from kafka is consumed. I think the offset should only change when the offset has been committed otherwise the behavior of the stream function will be inconsistent

dantswain:
I can publish 100 messages to a topic and create a stream for that topic:

```
iex(8)> Enum.each(1 .. 100, fn(ix) -> KafkaEx.produce("mytopic", 0, "Message #{ix}") end)
:ok

iex(9)> s = KafkaEx.stream("mytopic", 0, auto_commit: true, consumer_group: "test", max_bytes: 100)
%KafkaEx.Stream{consumer_group: "test",
 fetch_request: %KafkaEx.Protocol.Fetch.Request{auto_commit: true,
  client_id: nil, correlation_id: nil, max_bytes: 100, min_bytes: 1, offset: 0,
  partition: 0, topic: "mytopic", wait_time: 10}, no_wait_at_logend: false,
 worker_name: :kafka_ex}
```

Note it sets the offset to 0 because the consumer group is new.  Now I take 20 messages, which should be well over 100 bytes.  It actually ends up with duplicate messages because it's over the max_bytes:

```
iex(11)> m = Enum.take(s, 20)
[%KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1226351079, key: "",
  offset: 0, value: "Message 1"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3490836573, key: "",
  offset: 1, value: "Message 2"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3490836573, key: "",
  offset: 1, value: "Message 2"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 2803286219, key: "",
  offset: 2, value: "Message 3"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 2803286219, key: "",
  offset: 2, value: "Message 3"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 963794280, key: "",
  offset: 3, value: "Message 4"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 963794280, key: "",
  offset: 3, value: "Message 4"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1316316670, key: "",
  offset: 4, value: "Message 5"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1316316670, key: "",
  offset: 4, value: "Message 5"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3615241284, key: "",
  offset: 5, value: "Message 6"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 3615241284, key: "",
  offset: 5, value: "Message 6"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 2692416722, key: "",
  offset: 6, value: "Message 7"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 2692416722, key: "",
  offset: 6, value: "Message 7"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 818157891, key: "",
  offset: 7, value: "Message 8"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 818157891, key: "",
  offset: 7, value: "Message 8"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1203972565, key: "",
  offset: 8, value: "Message 9"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1203972565, key: "",
  offset: 8, value: "Message 9"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1394529882, key: "",
  offset: 9, value: "Message 10"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1394529882, key: "",
  offset: 9, value: "Message 10"},
 %KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 605677260, key: "",
  offset: 10, value: "Message 11"}]
```

If I take 1 again, it actually starts back at 0, because the `offset` value in my stream struct is 0:

```
iex(12)> m = Enum.take(s, 1)
[%KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 1226351079, key: "",
  offset: 0, value: "Message 1"}]
```

This makes sense if you look at the implementation of `KafkaEx.Stream` because it takes `s.offset` as the first offset to `fetch`.

This is even more fun:  If I create a NEW stream with the same consumer group, I get offset 2.  Furthermore, if I `take` one from that stream I start with message at offset 2, so I've actually completely skipped offset 1:

```
iex(13)> s2 = KafkaEx.stream("mytopic", 0, auto_commit: true, consumer_group: "test", max_bytes: 100)
%KafkaEx.Stream{consumer_group: "test",
 fetch_request: %KafkaEx.Protocol.Fetch.Request{auto_commit: true,
  client_id: nil, correlation_id: nil, max_bytes: 100, min_bytes: 1, offset: 2,
  partition: 0, topic: "mytopic", wait_time: 10}, no_wait_at_logend: false,
 worker_name: :kafka_ex}
iex(14)> m = Enum.take(s2, 1)
[%KafkaEx.Protocol.Fetch.Message{attributes: 0, crc: 2803286219, key: "",
  offset: 2, value: "Message 3"}]
```

I'm pretty sure can sort this all out.  The question is, from a design standpoint, should the stream continue along the topic from the last offset that was committed, or should it start at the same `offset`.  There are kind of conflicting ideas here - it's intuitive that subsequent calls would fetch the next messages, BUT this is a functional language and therefore there's an argument that making the same function call twice with the same arguments would return the same value.  I think we just need to make a decision either way.

Regardless, I definitely agree that documenting this is really important.

bjhaid:
Can you verify what the behavior of [`File.stream!`](https://hexdocs.pm/elixir/File.html#stream!/3) is, we should try to be consistent as much as possible with the read behavior of [`File.stream!`]

bjhaid:
nvm I just verified, and this is consistent with the behavior of `File.stream!`, returning duplicate elements is weird though, maybe we can try not doing that :)

dantswain:
`File.stream!` always starts at the same point in the file:

```
iex(20)> f = File.stream!("README.md")
%File.Stream{line_or_bytes: :line, modes: [:raw, :read_ahead, :binary],
 path: "README.md", raw: true}
iex(21)> Enum.take(f, 2)
["KafkaEx\n", "========\n"]
iex(22)> Enum.take(f, 2)
["KafkaEx\n", "========\n"]
```

dantswain:
I fixed the double-consume issue.  I think we may be committing in the wrong spot, so I want to add some tests for that.  Still need to update the documentation as well.

There's a potential weirdness with `auto_commit`.  Suppose `auto_commit` is true, we have 10 messages in the log, and someone does `Enum.take(stream, 3)`.  With smallish messages, it's entirely possible that the fetch will get us all 10 messages in the log, but the user is only consuming 3 of them.  I _think_ we can handle this because the `acc` argument to reduce will have the `take` argument, but I'm not sure if that's an intended usage of the API... I'll look into it.

dantswain:
@bjhaid @joshuawscott I've handled everything I could think of, and updated the documentation.  Could you guys give this another review, please?

One thing I considered doing was delegating stream creation to a function in the `KafkaEx.Stream` module, and moving the bulk of the `KafkaEx.stream/3` documentation to the moduledocs for `KafkaEx.Stream`.  Doing that could get a little tricky because `KafkaEx.stream/3` uses some private function inside the root module.

test/integration/consumer_group_test.exs
```diff
@@ -240,31 +240,156 @@ defmodule KafkaEx.ConsumerGroup.Test do
...
+      topic_name,
+      0,
+      worker_name: :stream,
+      offset: 0,
+      no_wait_at_logend: true,
+      auto_commit: true,
+      consumer_group: consumer_group
+    )
+
+    [m1, m2, m3, m4] = Enum.take(stream, 10)
```
joshuawscott: Worth testing that this actually only returns 4 elements?

test/integration/consumer_group_test.exs
```diff
@@ -240,31 +240,156 @@ defmodule KafkaEx.ConsumerGroup.Test do
...
+      0,
+      worker_name: :stream,
+      no_wait_at_logend: true,
+      auto_commit: true,
+      offset: 0,
+      consumer_group: consumer_group
+    )
+
+    assert ["Msg 1", "Msg 2"] == stream1
+    |> Enum.take(2)
```
joshuawscott: nitpicky, but I kind of prefer assigning to a variable first so that the assert is more self-contained
```
messages = stream1 
|> Enum.take(2) 
|> Enum.map(&(&1.value))
assert ["Msg 1", "Msg 2"] == messages
```
Mainly because I was super confused that a list == a stream before I noticed the pipeline on the next line.

test/integration/consumer_group_test.exs
dantswain: This would match fail if I didn't get 4 elements, or am I misunderstanding you?

joshuawscott: *goes off to drink more coffee before reviewing more*

test/integration/consumer_group_test.exs
```diff
@@ -240,31 +240,158 @@ defmodule KafkaEx.ConsumerGroup.Test do
...
+      ]
+    }, worker_name: :stream)
+
+    stream = KafkaEx.stream(
+      topic_name,
+      0,
+      worker_name: :stream,
+      offset: 0,
+      auto_commit: true,
+      consumer_group: "stream_test"
```
bjhaid: We should use the `consumer_group` var here instead of `"stream_test"`

Issue #94, opened 2016-03-20
Sync timeouts greater than 5 seconds are not useful


dantswain:
Calls that pass through KafkaEx.Server will time out at 5 seconds regardless of the setting of sync_timeout because of the GenServer.call's default timeout.

At first I thought we could just grab the value of sync_timeout and modify all of the GenServer.calls to have an appropriate timeout, but this gets messy if you ever want to support different timeout values to different KafkaEx.Server instances.

I can think of three basic solutions here:
- Support a single global (still configurable) value for sync timeout.  I think if you want to go this route we should remove the sync timeout value from the GenServer state b/c it isn't actually state anymore.
- Support a per-worker timeout value.  This would mean allowing the value to be passed into create_worker.  It would also make modifying the GenServer.calls to support higher timeout values tricky because the timeout value would have to be tailored to each process somehow.
- Add optional timeout arguments to API calls that interact with the GenServer processes.  We could optionally pass these timeouts through to NetworkClient.

The third option could be combined with either of the first two.  What do you think, @bjhaid?

bjhaid:
I like a combination of 1 and 3 where the value provided in option 3 overrides the value in 1

dantswain:
This should be closed as of #196

Issue #67, opened 2016-02-07
Update consumer group API to support kafka 0.9.0


bjhaid:
`ConsumerMetatdataRequest` was [removed](https://cwiki.apache.org/confluence/pages/diffpages.action?originalId=61329518&pageId=61330213) from the kafka 0.9.0 API, and replaced with a better API. We need to support the new API as described at https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol

dantswain:
@bjhaid Does this mean that kafka_ex will not work with a 0.9.0 broker?

dantswain:
OK, I think I misread.  Our reading of the upgrade docs says that there will be no breakage - you're just suggesting that the new consumer group API be integrated into kafka_ex.

For reference for anyone else:
http://docs.confluent.io/2.0.0/upgrade.html

> Pay careful consideration to the order in which components are upgraded. Kafka is backward compatible, which means that clients from 0.8.x releases (CP 1.0.x) will work with brokers from 0.9.x releases (CP 2.0.0), but not vice-versa. This means you always need to plan upgrades such that all brokers are upgraded before clients. Clients include any application that uses Kafka producer or consumer, command line tools, Camus, Schema Registry and Rest Proxy.

bjhaid:
@dantswain you are correct, our integration tests currently run against 0.9.0 without issues, kafka is generally backwards compatible, this issue is to support the new API introduced in 0.9.0

dantswain:
@bjhaid Do you have any plans to work on this / have started it?  I might be able to get on it during work hours since we're pushing to get rid of Poseidon (which is no longer maintained! yay!).

bjhaid:
Thanks for your interest, I am yet to start working on this, and as usual I will appreciate your contributions.

Thanks once again

dantswain:
@bjhaid How would you feel about breaking this down into bite-sized chunks?  I can probably get some hours to work on this, but it's a pretty big undertaking.  I was thinking it might be more manageable if I (or anyone else) can start by implementing structs for the new protocol messages and submitting PRs for those individually.  The down side is that it would be asking you to merge into master effectively useless code (implementation of structs but no integration of them), but that doesn't seem too horrible, especially if we maybe add disclaimers that the full feature is not yet supported.  Another option would be to create a feature branch and submit PRs against that.

bjhaid:
[...]

PR #195, opened 2017-03-04
Consumer Group

dcuddeback:
I've taken a shot at implementing the consumer group API (#67).  There's still some more to do for this PR, but I wanted to get some feedback on some specific questions before going further.  First, I'll explain the implementation.

At a high level, this PR provides an API that allows one to write a consumer like this:

```elixir
defmodule MyApp.Consumer do
  use KafkaEx.GenConsumer

  require Logger

  def init(topic, partition) do
    {:ok, {topic, partition}}
  end

  def handle_message(%Message{offset: offset, value: message}, {topic, partition} = state) do
    Logger.info("#{topic}/#{partition}@#{offset}: #{inspect(message)}")

    {:ack, state}
  end
end
```

The API is modeled after `GenServer`.  `init/2` is expected to return `{:ok, state}`.  `handle_message/2` is passed a message and state and is expected to return an instruction (`:ack`, `:commit`, etc) and a new state.

The consumer can be launched as a consumer group like this:

```elixir
{:ok, pid} = KafkaEx.ConsumerGroup.start_link(MyApp.Consumer, "group_name", ~w[topic1 topic2])
```

It's also possible to launch an individual consumer manually, if not using the consumer group protocol, but this is not the intended use case:

```elixir
{:ok, pid} = KafkaEx.GenConsumer.start_link(MyApp.Consumer, "group_name", "topic1", 0)
```

## Design

I think this is pretty close to what we brainstormed on Slack.

The implementation uses the `KafkaEx` module as a low-level API, which picks up four methods (`join_group/3`, `sync_group/5`, `leave_group/3`, and `heartbeat/4`), which implement the individual network calls.  The rest of the high-level consumer API is implemented as a layer on top of the interface provided by the `KafkaEx` module.

The [Kafka client-side assignment protocol](https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Client-side+Assignment+Proposal) is implemented by `KafkaEx.ConsumerGroup`.  This module manages an individual BEAM node's membership in a single consumer group.  After receiving a `SyncGroupResponse` from the broker, it launches a supervisor to manage the consumption from its assigned partitions.  When shutdown by its supervisor or notified by a Kafka broker that the consumer group needs to rebalance, it shuts down the process tree starting at `KafakEx.GenConsumer.Supervisor`, which triggers offset commits.

`KafkaEx.GenConsumer.Supervisor` is given a list of assigned partitions when it's started, which it uses to manage a group of `KafkaEx.GenConsumer` processes.  It launches one `GenConsumer` per assigned partition, as shown in the diagram above.

A single instance of `KafkaEx.GenConsumer` consumes from a single partition.  It requires an application to implement `handle_message/2`, which is called for each message consumed from the partition.  The implementation is required to return an atom to indicate how the message should be handled:

  * `:ack` means that the message was consumed successfully.  The highest acked offset will be committed at the next auto-commit point.  Auto-committing can be configured based on a number of acked messages or time since last commit.
  * `:commit` means that the message was consumed successfuly and the offset should be committed synchronously, before the next message is handled.  This would obviously hurt throughput, but could be used sparingly when a consumer makes non-idempotent changes, for example.

When shutdown, a `GenConsumer` will commit the last acked offset.

The `GenConsumer` will also be consulted by `ConsumerGroup` when determining partition assignments.  `GenConsumer.assign_partitions/2` implements round-robin assignment by default, but can be overridden by a client application to provide a different assignment strategy.

## Questions

While working on this, I ran into a few issues that I would like feedback on:

1. In the diagram above, there's a red box that says, "All requests serialized here." Each call to `GenServer.call()` is synchronous and will block until previous calls finish.  This is going to limit the available concurrency, because only one request can be in flight per `KafkaEx.Server` at any time, even though each `KafkaEx.Server` holds a connection to each broker.  All other `NetworkClient` connections owned by a `KafkaEx.Server` will be idle during a network request.

   This can be partially mitigated by having a pool of `KafkaEx.Server` workers and check out a worker for each call, but there's still a lot of active `NetworkClient` connections that can't be used concurrently.  It seems a more efficient option would be to pool `NetworkClient` connections on a per-broker basis.  Or maybe using `GenServer.reply()` to reply asynchronously instead of returning `{:reply, ...}` would be enough to allow multiple simultaneous calls.

   I'm curious what your thoughts are on this issue.  I don't feel like I know this library well enough yet to know the best way to approach this.

   Ultimately, I'm wondering what would be considered good enough for this PR that would be compatible with tackling this issue later, if possible.

2. My biggest stumbling block to understanding the library well is that I'm not sure what level of abstraction is intended for `KafkaEx.Server`.  The servers implement a lot of calls that map one-to-one to requests in the Kafka network protocol.  I added four more methods that follow that pattern and treated it as a low-level API interface for the rest of this PR.  Is that the intended way to use `KafkaEx.Server`?

   I see that it also has state for consumer groups and metadata as well as functionality for streaming.  Have you considered separating this functionality out into high-level modules and keeping `KafkaEx.Server` as a low-level interface for the network protocol?  That approach is what felt most natural to me in working on this PR.  `ConsumerGroup` is a separate `GenServer` which uses `KafakEx.Server` for its network protocol, but it feels like it's duplicating some of the state that's currently kept in the `KafkaEx.Server` implementations.

3. When shutting down `KafkaEx.GenConsumer.Supervisor`, I want to shutdown all of the supervised consumers in parallel, so that they can commit their offsets concurrently.  The only way I could find to get a supervisor to shutdown its children in parallel was to use the `:simple_one_for_one` restart strategy.  From the [Erlang docs](http://erlang.org/doc/man/supervisor.html):

   > As a `simple_one_for_one` supervisor can have many children, it shuts them all down asynchronously. This means that the children do their cleanup in parallel, and therefore the order in which they are stopped is not defined.

   `:simple_one_for_one` seems like it's geared more towards dynamically managed children, so the resulting code is a little messy IMO.  You can see the difference in commit [3aa8ae3](https://github.com/TheRealReal/kafka_ex/commit/3aa8ae33aee6b16fc1c2079102149e78242d2bf6).  I'm still learning OTP, so I'm wondering if you know a cleaner or more idiomatic way to shutdown a supervisor's children in parallel.

4. I haven't fleshed out all of the possible responses for `GenConsumer.handle_message/2`.  All the use cases I have are fine with just `:ack`, and `:commit` is a natural extension of `:ack`.

   I could use your help defining the initial return values.  I don't know enough about other people's use cases to know what's useful here.  It will be easier to add new return values than to change existing ones, so I'm erring on the side of fewer to start with.  Do you have any use cases that aren't satisfied by `:ack` and `:commit`?

   I could see adding `:noreply` with `GenConsumer.ack/2` and `GenConsumer.commit/2` methods (analogous to `GenServer.reply/2`) to allow offloading message handling to another process, such as a `GenStage`, but I don't have a use case to motivate it.

5. I made the network and `GenServer` timeout an option that can be passed to `KafkaEx.join_group/3` and friends.  This was necessary for `join_group`, because the broker doesn't respond until all group members join or the session timeout expires.  Since the default minimum for session timeouts (`group.min.session.timeout.ms`) is 6 seconds and `JoinGroupRequest` can potentially block for as long as the session timeout, this can easily exceed the default `GenServer` timeout of 5 seconds.

   There's a ticket related to this (#94).  If you like the way I handled this for `join_group/3`, `sync_group/5`, `leave_group/3`, and `heartbeat/4`, I could propogate the change to the rest of the `KafkaEx` methods that use `GenServer.call`.  Please let me know if you think that's the right direction.

### DIFF
--- a/lib/kafka_ex/consumer_group.ex
+++ b/lib/kafka_ex/consumer_group.ex
@@ -1,0 +1,309 @@
+defmodule KafkaEx.ConsumerGroup do
+  @moduledoc """
+  A process that manages membership in a Kafka consumer group.
+
+  Consumers in a consumer group coordinate with each other through a Kafka broker to distribute the
+  work of consuming one or several topics without any overlap. This is facilitated by the [Kafka
+  client-side assignment
+  protocol](https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Client-side+Assignment+Proposal).
+
+  Any time group membership changes (a member joins or leaves the group), a Kafka broker initiates
+  group synchronization by asking one of the group members (the leader) to provide partition
+  assignments for the whole group. Partition assignment is handled by the
+  `c:KafkaEx.GenConsumer.assign_partitions/2` callback of the provided consumer module.
+
+  A `ConsumerGroup` process is responsible for:
+
+  1. Maintaining membership in a Kafka consumer group.
+  2. Determining partition assignments if elected as the group leader.
+  3. Launching and terminating `GenConsumer` processes based on its assigned partitions.
+
+  To use a `ConsumerGroup`, a developer must define a module that implements the
+  `KafkaEx.GenConsumer` behaviour and start a `ConsumerGroup` with that module.
+
+  ## Example
+
+  The following consumer prints each message with the name of the node that's consuming the message:
+
+  ```
+  defmodule DistributedConsumer do
+    use KafkaEx.GenConsumer
+
+    def handle_message(%Message{value: message}, state) do
+      IO.puts(to_string(node()) <> ": " <> inspect(message))
+      {:ack, state}
+    end
+  end
+
+  # use DistributedConsumer in a consumer group
+  {:ok, pid} = KafkaEx.ConsumerGroup.start_link(DistributedConsumer, "test_group", ["test_topic"])
+  ```
+
+  Running this on multiple nodes might display the following:
+
+  ```txt
+  node1@host: "messages"
+  node2@host: "on"
+  node2@host: "multiple"
+  node1@host: "nodes"
+  ```
+
+  It is not necessary for the nodes to be connected, because `ConsumerGroup` uses Kafka's built-in
+  group coordination protocol.
+  """
+
+  use GenServer
+
+  alias KafkaEx.Config
+  alias KafkaEx.Protocol.JoinGroup.Request, as: JoinGroupRequest
+  alias KafkaEx.Protocol.Heartbeat.Request, as: HeartbeatRequest
+  alias KafkaEx.Protocol.Heartbeat.Response, as: HeartbeatResponse
+  alias KafkaEx.Protocol.LeaveGroup.Request, as: LeaveGroupRequest
+  alias KafkaEx.Protocol.LeaveGroup.Response, as: LeaveGroupResponse
+  alias KafkaEx.Protocol.Metadata.Response, as: MetadataResponse
+  alias KafkaEx.Protocol.Metadata.TopicMetadata
+  alias KafkaEx.Protocol.Metadata.PartitionMetadata
+  alias KafkaEx.Protocol.SyncGroup.Request, as: SyncGroupRequest
+  alias KafkaEx.Protocol.SyncGroup.Response, as: SyncGroupResponse
+
+  require Logger
+
+  defmodule State do
+    @moduledoc false
+    defstruct [
+      :worker_name,
+      :heartbeat_interval,
+      :session_timeout,
+      :consumer_module,
+      :consumer_opts,
+      :group_name,
+      :topics,
+      :partitions,
+      :member_id,
+      :generation_id,
+      :assignments,
+      :consumer_pid,
+    ]
+  end
+
+  @heartbeat_interval 5_000
+  @session_timeout 30_000
+
+  # Client API
+
+  @doc """
+  Starts a `ConsumerGroup` process linked to the current process.
+
+  This can be used to start a `ConsumerGroup` as part of a supervision tree.
+
+  `module` is a module that implements the `KafkaEx.GenConsumer` behaviour. `group_name` is the name
+  of the consumer group. `topics` is a list of topics that the consumer group should consume from.
+  `opts` can be any options accepted by `GenConsumer` or `GenServer`.
+
+  ### Return Values
+
+  This function has the same return values as `GenServer.start_link/3`.
+
+  If the consumer group is successfully created and initialized, this function returns `{:ok, pid}`,
+  where `pid` is the PID of the consumer group process.
+  """
+  @spec start_link(module, binary, [binary], KafkaEx.GenConsumer.options) :: GenServer.on_start
+  def start_link(consumer_module, group_name, topics, opts \\ []) do
+    {server_opts, consumer_opts} = Keyword.split(opts, [:debug, :name, :timeout, :spawn_opt])
+
+    GenServer.start_link(__MODULE__, {consumer_module, group_name, topics, consumer_opts}, server_opts)
+  end
+
+  # GenServer callbacks
+
+  def init({consumer_module, group_name, topics, opts}) do
+    worker_name = Keyword.get(opts, :worker_name, Config.default_worker)
+    heartbeat_interval = Keyword.get(opts, :heartbeat_interval, Application.get_env(:kafka_ex, :heartbeat_interval, @heartbeat_interval))
+    session_timeout = Keyword.get(opts, :session_timeout, Application.get_env(:kafka_ex, :session_timeout, @session_timeout))
+
+    consumer_opts = Keyword.drop(opts, [:heartbeat_interval, :session_timeout])
+
+    state = %State{
+      worker_name: worker_name,
+      heartbeat_interval: heartbeat_interval,
+      session_timeout: session_timeout,
+      consumer_module: consumer_module,
+      consumer_opts: consumer_opts,
+      group_name: group_name,
+      topics: topics,
+      member_id: "",
+    }
+
+    Process.flag(:trap_exit, true)
+
+    {:ok, state, 0}
+  end
+
+  def handle_info(:timeout, %State{generation_id: nil, member_id: ""} = state) do
+    new_state = join(state)
+
+    {:noreply, new_state, new_state.heartbeat_interval}
+  end
+
+  def handle_info(:timeout, %State{} = state) do
+    new_state = heartbeat(state)
+
+    {:noreply, new_state, new_state.heartbeat_interval}
+  end
+
+  def handle_info({:EXIT, pid, reason}, %State{consumer_pid: pid} = state) do
+    new_state = %State{state | consumer_pid: nil}
+
+    {:stop, reason, new_state}
+  end
+
+  def handle_info({:EXIT, _pid, _reason}, %State{} = state) do
+    {:noreply, state, state.heartbeat_interval}
+  end
+
+  def terminate(_reason, %State{} = state) do
+    leave(state)
+  end
+
+  # Helpers
+
+  defp join(%State{worker_name: worker_name, session_timeout: session_timeout, group_name: group_name, topics: topics, member_id: member_id} = state) do
+    join_request = %JoinGroupRequest{
+      group_name: group_name,
+      member_id: member_id,
+      topics: topics,
+      session_timeout: session_timeout,
+    }
+
+    join_response = KafkaEx.join_group(join_request, worker_name: worker_name, timeout: session_timeout + 5000)
+    new_state = %State{state | member_id: join_response.member_id, generation_id: join_response.generation_id}
+
+    Logger.debug("Joined consumer group #{group_name}")
+
+    if join_response.member_id == join_response.leader_id do
+      sync_leader(new_state, join_response.members)
+    else
+      sync_follower(new_state)
+    end
+  end
+
+  defp sync_leader(%State{worker_name: worker_name, topics: topics, partitions: nil} = state, members) do
+    %MetadataResponse{topic_metadatas: topic_metadatas} = KafkaEx.metadata(worker_name: worker_name)
+
+    partitions = Enum.flat_map(topics, fn (topic) ->
+      %TopicMetadata{error_code: :no_error, partition_metadatas: partition_metadatas} = Enum.find(topic_metadatas, &(&1.topic == topic))
+
+      Enum.map(partition_metadatas, fn (%PartitionMetadata{error_code: :no_error, partition_id: partition_id}) ->
+        {topic, partition_id}
+      end)
+    end)
+
+    sync_leader(%State{state | partitions: partitions}, members)
+  end
+
+  defp sync_leader(%State{worker_name: worker_name, session_timeout: session_timeout,
+                          group_name: group_name, generation_id: generation_id, member_id: member_id} = state, members) do
+    assignments = assign_partitions(state, members)
+
+    sync_request = %SyncGroupRequest{
+      group_name: group_name,
+      member_id: member_id,
+      generation_id: generation_id,
+      assignments: assignments,
+    }
+
+    sync_request
+    |> KafkaEx.sync_group(worker_name: worker_name, timeout: session_timeout + 5000)
+    |> update_assignments(state)
+  end
+
+  defp sync_follower(%State{worker_name: worker_name, session_timeout: session_timeout,
+                            group_name: group_name, generation_id: generation_id, member_id: member_id} = state) do
+    sync_request = %SyncGroupRequest{
+      group_name: group_name,
+      member_id: member_id,
+      generation_id: generation_id,
+      assignments: [],
+    }
+
+    sync_request
+    |> KafkaEx.sync_group(timeout: session_timeout + 5000, worker_name: worker_name)
+    |> update_assignments(state)
+  end
+
+  defp update_assignments(%SyncGroupResponse{error_code: :rebalance_in_progress}, %State{} = state), do: rebalance(state)
+  defp update_assignments(%SyncGroupResponse{error_code: :no_error, assignments: assignments}, %State{} = state) do
+    start_consumer(state, assignments)
+  end
+
+  defp heartbeat(%State{worker_name: worker_name, group_name: group_name, generation_id: generation_id, member_id: member_id} = state) do
+    heartbeat_request = %HeartbeatRequest{
+      group_name: group_name,
+      member_id: member_id,
+      generation_id: generation_id,
+    }
+
+    case KafkaEx.heartbeat(heartbeat_request, worker_name: worker_name) do
+      %HeartbeatResponse{error_code: :no_error} ->
+        state
+
+      %HeartbeatResponse{error_code: :rebalance_in_progress} ->
+        rebalance(state)
+    end
+  end
+
+  defp rebalance(%State{} = state) do
+    state
+    |> stop_consumer()
+    |> join()
+  end
+
+  defp leave(%State{worker_name: worker_name, group_name: group_name, member_id: member_id} = state) do
+    stop_consumer(state)
+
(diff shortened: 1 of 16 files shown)


briankereszturi:
I've been testing this branch with some data (a few hundred records/sec) and it's been working pretty well. Let me know if there's anything I can do to help get this merged.

dcuddeback:
@briankereszturi Thanks. I'm working through the issues reported by @ebert. I've written much of the documentation already. The README could use an update, and I think it could use some tests, too. I could use a hand with either of those.

briankereszturi:
I'll take a crack at those within the next day or two.

dcuddeback:
Rebased to incorporate the changes from #196.

dantswain:
@dcuddeback Thanks for your patience (or at least hiding your impatience ;) ).
I'm traveling this week and ironically I should have some time to give this the
detailed look it deserves (airports, hotels, etc).

I'll start with responding to your questions before diving into the code.
@bjhaid can feel free to disagree or chime in with a different perspective.

> 1. In the diagram above, there's a red box that says, "All requests serialized here." Each call to `GenServer.call()` is synchronous and will block until previous calls finish.  This is going to limit the available concurrency, because only one request can be in flight per `KafkaEx.Server` at any time, even though each `KafkaEx.Server` holds a connection to each broker.  All other `NetworkClient` connections owned by a `KafkaEx.Server` will be idle during a network request.
> 
>    This can be partially mitigated by having a pool of `KafkaEx.Server` workers and check out a worker for each call, but there's still a lot of active `NetworkClient` connections that can't be used concurrently.  It seems a more efficient option would be to pool `NetworkClient` connections on a per-broker basis.  Or maybe using `GenServer.reply()` to reply asynchronously instead of returning `{:reply, ...}` would be enough to allow multiple simultaneous calls.
> 
>    I'm curious what your thoughts are on this issue.  I don't feel like I know this library well enough yet to know the best way to approach this.
> 
>    Ultimately, I'm wondering what would be considered good enough for this PR that would be compatible with tackling this issue later, if possible.

I think the cleanest solution _right now_ is to use one KafkaEx.Server per
partition.  That's generally what I do in production with "manual" partition
assignments.  This is definitely not optimal but I think finding a better
solution is outside the scope of this PR.

Connection pools may be something we need to do eventually, and I think
we may be able to get some good use out of fishcakez/connection
 
> 2. My biggest stumbling block to understanding the library well is that I'm not sure what level of abstraction is intended for `KafkaEx.Server`.  The servers implement a lot of calls that map one-to-one to requests in the Kafka network protocol.  I added four more methods that follow that pattern and treated it as a low-level API interface for the rest of this PR.  Is that the intended way to use `KafkaEx.Server`?
> 
>    I see that it also has state for consumer groups and metadata as well as functionality for streaming.  Have you considered separating this functionality out into high-level modules and keeping `KafkaEx.Server` as a low-level interface for the network protocol?  That approach is what felt most natural to me in working on this PR.  `ConsumerGroup` is a separate `GenServer` which uses `KafakEx.Server` for its network protocol, but it feels like it's duplicating some of the state that's currently kept in the `KafkaEx.Server` implementations.

Your observation here is astute.  KafkaEx is really generally a low-level
library, but I'm not sure we've done a great job of making that clear in the
documentation.  Without having read the code yet, it sounds like your usage of
KafkaEx.Server is as-intended.  It more-or-less is responsible for maintaining
the connection state and translating the KafkaEx API to/from the actual Kafka
protocol API.

When I go through the code I'll try to keep this in mind.

> 3. When shutting down `KafkaEx.GenConsumer.Supervisor`, I want to shutdown all of the supervised consumers in parallel, so that they can commit their offsets concurrently.  The only way I could find to get a supervisor to shutdown its children in parallel was to use the `:simple_one_for_one` restart strategy.  From the [Erlang docs](http://erlang.org/doc/man/supervisor.html):
> 
>    > As a `simple_one_for_one` supervisor can have many children, it shuts them all down asynchronously. This means that the children do their cleanup in parallel, and therefore the order in which they are stopped is not defined.
> 
>    `:simple_one_for_one` seems like it's geared more towards dynamically managed children, so the resulting code is a little messy IMO.  You can see the difference in commit [3aa8ae3](https://github.com/TheRealReal/kafka_ex/commit/3aa8ae33aee6b16fc1c2079102149e78242d2bf6).  I'm still learning OTP, so I'm wondering if you know a cleaner or more idiomatic way to shutdown a supervisor's children in parallel.


What is your motivation to kill the child workers in parallel?  Depending on
how the connections are managed they may need to be blocked on network IO
anyways, and I doubt the shutdown time will be particularly important on the
scale of any actual work being done.

If they do need to handle cleanup in parallel, one solution might be to use a
set of parallel tasks that send a plain-old `GenServer.call`, then `Task.await`
before actually killing the consumers.

This actually sounds like a pretty good use case for `:simple_one_for_one` if I
understand your description correclty.  I imagine there being a supervisor for
the consumer group that dynamically manages children as partitions are
assigned/reassigned.

> 4. I haven't fleshed out all of the possible responses for `GenConsumer.handle_message/2`.  All the use cases I have are fine with just `:ack`, and `:commit` is a natural extension of `:ack`.
> 
>    I could use your help defining the initial return values.  I don't know enough about other people's use cases to know what's useful here.  It will be easier to add new return values than to change existing ones, so I'm erring on the side of fewer to start with.  Do you have any use cases that aren't satisfied by `:ack` and `:commit`?
> 
>    I could see adding `:noreply` with `GenConsumer.ack/2` and `GenConsumer.commit/2` methods (analogous to `GenServer.reply/2`) to allow offloading message handling to another process, such as a `GenStage`, but I don't have a use case to motivate it.

I'm a little torn on `:ack` and `:commit`.  `:ack` is kind of an overloaded
term in message broker systems - so at first I was really confused about the
difference between `:ack` and `:commit`.  For example, in RabbitMQ, "ack"
actually means something similar to what "commit" means in Kafka (the message
is officially removed from the "queue" once it is acked).

Is there a use case for wanting to have a consumer be able to decide, on a
per-message basis, whether or not to commit the offset?  That is as opposed to
configuring that as an option to the consumer initially.

The only other return value I can think of would be some kind of `:error`
mechanism.  I'm not sure what actions would be taken for an error response,
though.  The supervisor could restart the worker, but I wonder if it would be
better to leave that up to the user to simply crash their worker on failure so
that the supervisor restarts it.

> 5. I made the network and `GenServer` timeout an option that can be passed to `KafkaEx.join_group/3` and friends.  This was necessary for `join_group`, because the broker doesn't respond until all group members join or the session timeout expires.  Since the default minimum for session timeouts (`group.min.session.timeout.ms`) is 6 seconds and `JoinGroupRequest` can potentially block for as long as the session timeout, this can easily exceed the default `GenServer` timeout of 5 seconds.
> 
>    There's a ticket related to this (#94).  If you like the way I handled this for `join_group/3`, `sync_group/5`, `leave_group/3`, and `heartbeat/4`, I could propogate the change to the rest of the `KafkaEx` methods that use `GenServer.call`.  Please let me know if you think that's the right direction.

Check out #196 - I haven't gotten into the code yet but hopefully that's
a similar solution (we should be consistent).

lib/kafka_ex/consumer_group.ex
```diff
@@ -0,0 +1,309 @@
...
+    %State{state | consumer_pid: nil}
+  end
+
+  defp assign_partitions(%State{consumer_module: consumer_module, partitions: partitions}, members) do
+    assignments =
+      consumer_module.assign_partitions(members, partitions)
+      |> Enum.map(fn ({member, topic_partitions}) ->
+        assigns =
+          topic_partitions
+          |> Enum.group_by(&(elem(&1, 0)), &(elem(&1, 1)))
```
dantswain: This is a bit dense - could it be split to a function that has a name indicating what it's doing?  Maybe just `|> Enum.map(fn({k, v}) -> {k, [v]} end)` (if I'm reading correctly)

dcuddeback:
@dantswain Thanks for the detailed feedback. Unfortunately, our availability isn't overlapping very well on this PR. As of yesterday, my focus is on moving my family to another state. I expect to be preoccupied with that through May. I'm still interested in finishing this PR. I just want you to be aware of my preoccupation as it will impact my responsiveness.

> I think the cleanest solution right now is to use one KafkaEx.Server per partition. That's generally what I do in production with "manual" partition assignments. This is definitely not optimal but I think finding a better solution is outside the scope of this PR.

That makes sense to me. How do you see the server processes being managed in this case? The simplest way I can think of is to start and link a server process to a `GenConsumer` each time a `GenConsumer` is started. That incurs some overhead when rebalancing since the connections wouldn't be reused. Other solutions that I can think of would require some kind of pooling mechanism, which means either adding a dependency or reinventing something like poolboy.

> KafkaEx is really generally a low-level library, [...]. Without having read the code yet, it sounds like your usage of KafkaEx.Server is as-intended.

Thanks for confirming this. My biggest worry was that my approach might be completely off the mark in this regard.

> What is your motivation to kill the child workers in parallel?

Just that the delay was noticeable when shutting down one-by-one. The consumer processes are shutdown before rejoining the group on a rebalance. The time to rejoin a group is potentially the heartbeat interval plus the time to shutdown consumers and commit their offsets. I wanted to reduce the chance that this could encroach on the session timeout.

> I'm a little torn on :ack and :commit. :ack is kind of an overloaded term in message broker systems - so at first I was really confused about the difference between :ack and :commit. For example, in RabbitMQ, "ack" actually means something similar to what "commit" means in Kafka 
(the message is officially removed from the "queue" once it is acked).

This honestly felt like the least well-defined part of the feature to me. I don't feel strongly about keeping those names. I had also though of calling `:commit` `:checkpoint`. Perhaps a better name for `:ack` would be `:ok`. That feels a little more idiomatic Elixir to me: `{:ok, new_state}`.

> Is there a use case for wanting to have a consumer be able to decide, on a per-message basis, whether or not to commit the offset? That is as opposed to configuring that as an option to the consumer initially.

I'm actually not sure if you're asking about the use case of choosing between `:ack` and `:commit` or something like `:ack` and `:error`. I guess that proves your point about "ack" and "commit" being confusing. I think you're asking about choosing between `:ack` and `:commit`, in which case the use case is speculative. I was thinking that having a way to commit synchronously would be helpful if consuming a heterogenous queue where some messages are more expensive to replay than others. It's not a use case that I have, but it was a natural extension of the auto-committing feature. Now that I verbalize it, it doesn't seem like a use case that would come up often.

> The only other return value I can think of would be some kind of :error mechanism. I'm not sure what actions would be taken for an error response, though. The supervisor could restart the worker, but I wonder if it would be better to leave that up to the user to simply crash their worker on failure so that the supervisor restarts it.

I was thinking along the lines of retry in X seconds with exponential back-off, but that could also be handled inside the consumer.

Having discussed the above three points regarding the return value for `handle_message`, how do you feel about simplifying it to just `{:ok, new_state}` and iterating when new use cases are discovered? It'll always be easier to add return values than to remove them.

> Check out #196 - I haven't gotten into the code yet but hopefully that's a similar solution (we should be consistent).

I saw that PR come in and already rebased to be consistent with it. I did add a parameter to allow overriding `sync_timeout` when calling `join_group`, `sync_group`, etc, because by design those API calls can block for as long as the session timeout. (https://github.com/kafkaex/kafka_ex/pull/195/files#diff-5489fae6787d83003bb4de65cf092c17L457)

lib/kafka_ex/consumer_group.ex
```diff
@@ -0,0 +1,309 @@
...
+      :topics,
+      :partitions,
+      :member_id,
+      :generation_id,
+      :assignments,
+      :consumer_pid,
+    ]
+  end
+
+  @heartbeat_interval 5_000
```
dcuddeback: Makes sense.

lib/kafka_ex/consumer_group.ex
```diff
@@ -0,0 +1,309 @@
...
+
+    Process.flag(:trap_exit, true)
+
+    {:ok, state, 0}
+  end
+
+  def handle_info(:timeout, %State{generation_id: nil, member_id: ""} = state) do
+    new_state = join(state)
+
+    {:noreply, new_state, new_state.heartbeat_interval}
```
dcuddeback: Thanks for the suggestion. That seems much cleaner than using `GenServer` timeouts.

lib/kafka_ex/consumer_group.ex
```diff
@@ -0,0 +1,309 @@
...
+  end
+
+  def handle_info({:EXIT, pid, reason}, %State{consumer_pid: pid} = state) do
+    new_state = %State{state | consumer_pid: nil}
+
+    {:stop, reason, new_state}
+  end
+
+  def handle_info({:EXIT, _pid, _reason}, %State{} = state) do
```
dcuddeback: The previous clause uses `pid` twice so it only matches when `pid` matches `consumer_pid` (the `GenConsumer.Supervisor`). I think this clause was matching exit signals from the supervisor that was shutdown gracefully during a rebalance (because a new supervisor is assigned to `consumer_pid`).

lib/kafka_ex/consumer_group.ex
```diff
@@ -0,0 +1,309 @@
...
+    Logger.debug("Joined consumer group #{group_name}")
+
+    if join_response.member_id == join_response.leader_id do
+      sync_leader(new_state, join_response.members)
+    else
+      sync_follower(new_state)
+    end
+  end
+
+  defp sync_leader(%State{worker_name: worker_name, topics: topics, partitions: nil} = state, members) do
```
dcuddeback: Sure.

lib/kafka_ex/consumer_group.ex
```diff
@@ -0,0 +1,309 @@
...
+
+    sync_request = %SyncGroupRequest{
+      group_name: group_name,
+      member_id: member_id,
+      generation_id: generation_id,
+      assignments: assignments,
+    }
+
+    sync_request
+    |> KafkaEx.sync_group(worker_name: worker_name, timeout: session_timeout + 5000)
```
dcuddeback: Makes sense.

lib/kafka_ex/consumer_group.ex
dcuddeback: I'll see what I can do to make it easier to follow. This is transforming a list of partitions into the format expected by `SyncGroupRequest`. That is `[{topic1, part1}, {topic1, part2}, {topic2, part1}]` gets transformed into `[{topic1, [part1, part2], {topic2, part1}]`.

dcuddeback:
> What do you think about this as a possible alternate structure wrt the supervision tree?
> 
> ConsumerGroup.start_link starts a supervisor, the supervisor has these children:
> 
> a) the ConsumerGroup worker you have here (some care will have to go into naming :/ )
> b) a ConsumerGroup.WorkerSupervisor that has :simple_one_for_one and child specs for the actual consumer workers
> 
> Then, on join/rebalance, we reap/restart the consumer workers. That way the supervision tree is all started at once and we don't have to have any processes dangling outside the supervision tree.

I didn't mean to have a process outside of the supervision tree. This is probably my misunderstanding of OTP. I thought that starting the supervisor with `start_link` from a supervised process made it part of the supervision tree. I'm curious what I'm missing.

I like your suggested structure, too. I'm not partial to starting/stopping the supervisor on rebalance. I modeled it as starting a supervisor with a list of partition assignments, because I started with `:one_for_one` which likes to know its children up front. It does seem like `:simple_one_for_one` would allow for the supervisor to be started ahead of time. Since I've already changed the supervisor to `:simple_one_for_one`, it's probably not too difficult to restructure it as you suggested.

dantswain:
@dcuddeback Do you think you'll have time to revisit this?  I'd hate for all the effort to go to waste, so if you don't think you'll have time I want to make sure we mark it as something someone else could pick up and make contributions to.

dcuddeback:
@dantswain I'd love to. I was working on this for my former employer, but I'm on a break now for a couple of months while I move out of state. My move-in date is in about three weeks. I could use a break from packing boxes, so maybe I can set aside a day next week to work on this.

dcuddeback:
Closing in favor of #207.

PR #198, opened 2017-03-21
Add missing nil case for socket info

mustafaturan:
Spec documentation has `nil` case for `socket` `info` function but the code and the test do not cover nil case `socket` variable.


dantswain:
@mustafaturan any plans to address the requested changes?

mustafaturan:
@dantswain, I might not have time in two months.
Feel free to close the PR.

Thanks for the great work, all the best.

dantswain:
Covered by #213. Thanks again @mustafaturan for bringing this up!

PR #203, opened 2017-04-25
Initialize consumer group metadata

### LINKED ISSUE #202: Consumer group metadata is uninitialized
When running in a clustered environment a `:not_coordinator_for_consumer` is thrown on any request before the `consumer_group_update_interval` elapses. It looks like the initial state is never initialized. 

```
  5) test can heartbeat (KafkaEx.Server0P9P0.Test)
     test/integration/server0_p_9_p_0_test.exs:92
     Assertion with == failed
     code:  answer.error_code() == :no_error
     left:  :not_coordinator_for_consumer
     right: :no_error
     stacktrace:
       test/integration/server0_p_9_p_0_test.exs:105: (test)
```

zapient:
Fixes #202 

Initialize consumer group metadata on startup. No tests were added because I was actually seeing `:not_coordinator_for_consumer` before my changes.

### DIFF
--- a/lib/kafka_ex/server_0_p_8_p_2.ex
+++ b/lib/kafka_ex/server_0_p_8_p_2.ex
@@ -55,10 +55,15 @@ defmodule KafkaEx.Server0P8P2 do
     state = update_metadata(state)
     {:ok, _} = :timer.send_interval(state.metadata_update_interval, :update_metadata)
 
-    # only start the consumer group update cycle if we are using consumer groups
-    if consumer_group?(state) do
-      {:ok, _} = :timer.send_interval(state.consumer_group_update_interval, :update_consumer_metadata)
-    end
+    state =
+      if consumer_group?(state) do
+        # If we are using consumer groups then initialize the state and start the update cycle
+        {_, updated_state} = update_consumer_metadata(state)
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/server_0_p_8_p_2.ex
```diff
@@ -55,10 +55,15 @@ defmodule KafkaEx.Server0P8P2 do
     state = update_metadata(state)
     {:ok, _} = :timer.send_interval(state.metadata_update_interval, :update_metadata)
 
-    # only start the consumer group update cycle if we are using consumer groups
-    if consumer_group?(state) do
-      {:ok, _} = :timer.send_interval(state.consumer_group_update_interval, :update_consumer_metadata)
-    end
+    state =
```
dantswain: @bjhaid or @cdegroot do you know if this feature is supported on 0.8.2?  I don't think it is, but was hoping one of you would know off-hand.

dantswain: Scratch that, I think I was misreading this.

zapient:
> I I'd like to see it factored so that there's not so much repeated code, but I think that's a problem with a much bigger scope.

Agreed. `kafka_server_init` seems ripe for refactoring but out of scope for this particular issue.

Issue #201, opened 2017-04-07
Start worker without kafka supervisor


steffenix:
Using the example provided on the readme
```
GenServer.start_link(KafkaEx.Server,
  [
    [uris: Application.get_env(:kafka_ex, :brokers),
     consumer_group: Application.get_env(:kafka_ex, :consumer_group)],
    :no_name
  ]
)
```

I get the error:

``` 
(UndefinedFunctionError) function KafkaEx.Server.init/1 is undefined or private

(kafka_ex) KafkaEx.Server.init([[uris: [{"10.10.101.1", 9092}, {"10.10.101.2", 9092}, {"10.10.101.3", 9092}], consumer_group: "kafka_ex_messenger"], :no_name]) 
```

I am using the latest version of kafka_ex 0.6.5.

----

After investigating, I can start it using KafkaEx.Server0P8P2 instead of KafkaEx.Server.

dantswain:
Thanks, @steffenix !

joshuawscott:
Seems to be fixed by #204

PR #204, opened 2017-05-09
Improved the example for creating a worker manually, so it doesn't re…

thuey:
…sult in an undefined error

See kafkaex/kafka_ex#201


dantswain:
:+1: LGTM.  I'll pull this and #203 together sometime over the next couple days and do a hotfix release.

WRT the wrapper - that's a good question and deserves some discussion.

PR #207, opened 2017-05-27
Consumer Group

dcuddeback:
This is a continuation of #195, since I don't have access to push to the branch for that PR. The changes in this PR address some of the feedback from #195. The only thing that I didn't address yet is the `:ack` and `:commit` responses. Changing those will require rewriting some of the documentation, so I'd like to agree on what the responses should be before doing that.


dantswain:
Closing in favor of #219

PR #211, opened 2017-07-07
Fix dialyzer issues

joshuawscott:
This also cleans up a naming conflict with sync_timeout. I renamed the
function that pulls from the config to config_sync_timeout to resolve
this.


test/test_helper.exs
```diff
@@ -4,8 +4,7 @@ ExUnit.configure exclude: [integration: true, consumer_group: true, server_0_p_9
 
 defmodule TestHelper do
   def generate_random_string(string_length \\ 20) do
-    :random.seed(:os.timestamp)
-    Enum.map(1..string_length, fn _ -> (:random.uniform * 25 + 65) |> round end) |> to_string
+    Enum.map(1..string_length, fn _ -> (:rand.uniform * 25 + 65) |> round end) |> to_string
```
bjhaid: we need to seed the random number generator

bjhaid:
@joshuawscott this seems like a straight forward PR, other than the seeding of the RNG, I can almost click the merge button.

test/test_helper.exs
joshuawscott: :rand.uniform doesn't need seeding according to the docs http://erlang.org/doc/man/rand.html

bjhaid: thanks

bjhaid: hey so it seems `:rand.uniform` isn't available in otp 17, should we have a module/function to decorate this, or kill support for otp 17

joshuawscott: We use OTP 19 + elixir 1.4 everywhere right now. I can certainly put :random.uniform back if we want to keep supporting OTP 17

bjhaid: @joshuawscott I don't know if we should be supporting it or not, you can make the call if you think we shouldn't support it please delete it here https://github.com/kafkaex/kafka_ex/blob/master/.travis.yml#L18-L19 and I'll merge

PR #213, opened 2017-07-11
Refactor broker selection logic

dantswain:
This is motivated by #198 and the discussion around it.  My hope is that
this makes the specs correct and the code a little easier to follow.

### DIFF
--- a/lib/kafka_ex/protocol/metadata.ex
+++ b/lib/kafka_ex/protocol/metadata.ex
@@ -1,5 +1,4 @@
 defmodule KafkaEx.Protocol.Metadata do
-  alias KafkaEx.Socket
   alias KafkaEx.Protocol
   import KafkaEx.Protocol.Common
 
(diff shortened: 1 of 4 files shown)


lib/kafka_ex/protocol/metadata.ex
```diff
@@ -39,28 +51,13 @@ defmodule KafkaEx.Protocol.Metadata do
...
 
-    defp broker_with_open_socket(broker) do
-      case Socket.info(broker.socket) do
-        port_info when is_list(port_info) -> broker
-        _ -> nil
-      end
-    end
-  end
-
-  defmodule Broker do
```
dantswain: Github diff is making this more complicated - I had to move the Broker module towards the top to get the compiler to work with the ordering.

lib/kafka_ex/socket.ex
```diff
@@ -79,11 +79,19 @@ defmodule KafkaEx.Socket do
   end
 
   @doc """
+  Returns true if the socket is open
+  """
+  @spec open?(KafkaEx.Socket.t) :: boolean
+  def open?(socket = %KafkaEx.Socket{}) do
+    info(socket) != nil
```
joshuawscott: The spec says `info/1` only returns `list`, so would this always be true?

dantswain: Good catch.  `Port.info` returns `nil` if the port is closed, so I just had the spec for `Socket.info/2` incorrect.

PR #219, opened 2017-07-25
Consumer group (Again)

dantswain:
This is my continuation of #207 which is in turn a continuation of #195 .

It's still WIP - feel free to review as it goes, but I'll post when done.


dantswain:
@bjhaid @joshuawscott I think this is finally ready for review.  There are a couple things I might like to clean up later, but they can be in follow-up PRs.  I mostly wanted to nail down the API first.

@dcuddeback If you have time, I would greatly appreciate any review you can give here.  This is still mostly your vision, but I changed a handful of things.  This may not be an exhaustive list:

* Renamed `:ack` / `:commit` to `:async_commit` and `:sync_commit`, respectively.
* `GenConsumer` handles message _sets_ instead of individual messages now.  @joshuawscott and I have found this to be useful in some applications (e.g., we can process messages in parallel), and it's trivial for a user to iterate over the message set on their own if they want to.
* `ConsumerGroup` is now the supervisor and it starts a `ConsumerGroup.Manager` process.  I thought this made more sense from an API standpoint - `ConsumerGroup.start_link` is now the fundamental entry point for a consumer group.
* Remove partition assignment responsibility from `GenConsumer` - this isn't really part of consuming messages.  Round-robin is probably fine for most users.  Users still have the ability to specify a partition assignment callback in `ConsumerGroup.start_link`.
* Gave the documentation a relatively thorough editing.
* Added tests for the consumer group.
* Added a bunch of ops-friendly functions.

bjhaid:
ack, this is some massive diff, I'll find sometime to review it before the end of the week

lib/kafka_ex/consumer_group/manager.ex
```diff
@@ -0,0 +1,476 @@
...
+    } = state
+  ) do
+    join_request = %JoinGroupRequest{
+      group_name: group_name,
+      member_id: member_id || "",
+      topics: topics,
+      session_timeout: session_timeout,
+    }
+
+    join_response = %JoinGroupResponse{error_code: :no_error} =
```
joshuawscott: This will crash if we get any error back from Kafka. I don't know if it's necessarily good to crash within a private function that's in an essentially private module. It will make it difficult to diagnose for a library user. Maybe we should crash with a helpful message here (like display a human-readable version of the error code that we got)

lib/kafka_ex/gen_consumer.ex
```diff
@@ -0,0 +1,612 @@
...
+  Any other return value will cause the `start_link/5` to return `{:error,
+  error}` and the process to exit.
+  """
+  @callback init(topic :: binary, partition :: non_neg_integer) ::
+    {:ok, state :: term}
+
+  @doc """
+  Invoked for each message set consumed from a Kafka topic partition.
+
+  `message_set` is a message set fetched from a Kafka broker and `state` is the
```
joshuawscott: maybe say "list" instead of "set", or even just "Enumerable", otherwise it sort of sounds like it's a MapSet or something.

lib/kafka_ex/gen_consumer.ex
```diff
@@ -0,0 +1,612 @@
...
+  """
+  @callback handle_call(call :: term, from :: GenServer.from, state :: term)
+    :: {:reply, reply_value :: term, new_state :: term}
+
+  defmacro __using__(_opts) do
+    quote do
+      @behaviour KafkaEx.GenConsumer
+      alias KafkaEx.Protocol.Fetch.Message
+
+      def init(_topic, _partition) do
```
joshuawscott: It seems like we should put the topic and partition into the state by default?

lib/kafka_ex/gen_consumer.ex
```diff
@@ -0,0 +1,612 @@
...
+        {:ok, nil}
+      end
+
+      def handle_call(_call, _from, _consumer_state) do
```
joshuawscott: I didn't see anywhere that it's explicitly called out in the documentation which calls should be handled when this is implemented - we call the `:partition` and `{:consumer_call, _}` forms from this module, so it seems like we should either spell that out, or implement a default callback.

joshuawscott: We could just remove this, we already define a handle_call, and it should be obvious that you have to define handle_call to make a GenServer.call against a process.

test/integration/consumer_group_implementation_test.exs
```diff
@@ -0,0 +1,277 @@
...
+        assert ConsumerGroup.leader?(context[:consumer_group_pid2])
+      true ->
+        raise "Neither member is the leader"
+    end
+
+    assignments1 = ConsumerGroup.assignments(context[:consumer_group_pid1])
+    assignments2 = ConsumerGroup.assignments(context[:consumer_group_pid2])
+    assert 2 == length(assignments1)
+    assert 2 == length(assignments2)
+    refute assignments1 == assignments2
```
joshuawscott: maybe assert that none of the assignments in `assignments1` are in `assignments2`

lib/kafka_ex/consumer_group/manager.ex
dantswain: I think what I'd like to do here is follow up on some of these with separate PRs.  I'll make issues for them.

This one is #223

dantswain:
@bjhaid @dcuddeback ping.  Sorry for the size of the PR :(

PR #222, opened 2017-08-01
Updates to CI - drops testing OTP 17

dantswain:
* Only re-run the tests if necessary (in the new consumer-group branch this is not usually necessary)
* Drop building/testing Elixir 1.1.1 with OTP 17
* Update OTP 19.2 to OTP 19.3
* Update Elixir 1.4.0 to 1.4.5
* Add Elixir 1.5.0 with OTP 19.3 and OTP 20.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,24 +8,26 @@ cache:
     - _build
 otp_release:
   - 18.3
-  - 19.2
+  - 19.3
 elixir:
-  - 1.4.0
+  - 1.4.5
   - 1.3.4
   - 1.2.6
   - 1.1.1
 matrix:
   exclude:
     - elixir: 1.1.1
-      otp_release: 19.2
-    - elixir: 1.4.0
-      otp_release: 19.2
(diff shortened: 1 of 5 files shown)


bjhaid:
@dantswain can you please remove the logic added in this commit:

https://github.com/kafkaex/kafka_ex/commit/8038a2fb062ca2d9a08ee283ee06fd7545198d83

it's to accommodate OTP 17 in the tests

dantswain:
@bjhaid Done.  I've also updated the README and updated mix.exs to indicate we require at least Elixir 1.1.

bjhaid:
Should we come up with some policy on deprecating support for older versions of OTP and elixir?

dantswain:
I'm not a big fan of policies, but 3 versions behind the newest tested version seems reasonable?  Though by that logic we could also drop Elixir 1.1.  I'm mostly just dropping R17 here because it's becoming a maintenance burden.

PR #231, opened 2017-09-18
Use 'iolist' when building packet.

vjache:
This change makes using iolist instead of contiguous binary when building packet being compressed and/or sent.

P.S. Unit tests & dialyzer passed successfully but I'm failed to wait for finish integration tests, they just hanging without any log messages. Anyway, I making this PR, in case someone find time to run all integration tests.

### DIFF
--- a/lib/kafka_ex/protocol/produce.ex
+++ b/lib/kafka_ex/protocol/produce.ex
@@ -32,41 +32,45 @@ defmodule KafkaEx.Protocol.Produce do
   end
 
   def create_request(correlation_id, client_id, %Request{topic: topic, partition: partition, required_acks: required_acks, timeout: timeout, compression: compression, messages: messages}) do
-    message_set = create_message_set(messages, compression)
-    KafkaEx.Protocol.create_request(:produce, correlation_id, client_id) <>
-      << required_acks :: 16-signed, timeout :: 32-signed, 1 :: 32-signed >> <>
-      << byte_size(topic) :: 16-signed, topic :: binary, 1 :: 32-signed, partition :: 32-signed, byte_size(message_set) :: 32-signed >> <> message_set
+    {message_set, mssize} = create_message_set(messages, compression)
+    [KafkaEx.Protocol.create_request(:produce, correlation_id, client_id),
+      << required_acks :: 16-signed, timeout :: 32-signed, 1 :: 32-signed >>,
+      << byte_size(topic) :: 16-signed, topic :: binary, 1 :: 32-signed, partition :: 32-signed, mssize :: 32-signed >>,
+      message_set]
   end
 
   def parse_response(<< _correlation_id :: 32-signed, num_topics :: 32-signed, rest :: binary >>), do: parse_topics(num_topics, rest, __MODULE__)
   def parse_response(unknown), do: unknown
 
-  defp create_message_set([], _compression_type), do: ""
+  defp create_message_set([], _compression_type), do: {"", 0}
   defp create_message_set([%Message{key: key, value: value}|messages], :none) do
-    message = create_message(value, key)
-    message_set = << 0 :: 64-signed >> <> << byte_size(message) :: 32-signed >> <> message
-    message_set <> create_message_set(messages, :none)
+    {message, msize} = create_message(value, key)
+    message_set = [<< 0 :: 64-signed >>, << msize :: 32-signed >>, message]
+    {message_set2, ms2size} = create_message_set(messages, :none)
+    {[message_set, message_set2], 8 + 4 + msize + ms2size}
   end
   defp create_message_set(messages, compression_type) do
-    message_set = create_message_set(messages, :none)
+    {message_set, _} = create_message_set(messages, :none)
     {compressed_message_set, attribute} =
       Compression.compress(compression_type, message_set)
-    message = create_message(compressed_message_set, nil, attribute)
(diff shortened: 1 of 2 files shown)


joshuawscott:
Thanks for the PR!

Travis ran the integration tests successfully, so all looks good there. I'm unsure of the reasoning behind the change, though, could you give a rundown of the advantages of doing it this way instead of with binaries?

Cheers!

bjhaid:
I think there are claims that iolist performs better when constructing data, but I believe the claims are mostly true when constructing data from a combination of charlists and binary. I'll like to see benchmarks that demonstrates that this is an improvement over what we currently have. And if my guess is wrong a rationale around why we should use iolist in lieu of the current implementation.

vjache:
The matter is not that iolists claimed faster or slower. The key point here is that KafkaEx is just a kind of transport for message transmitting and those messages often produced by some serialization libraries and those libraries when construct a binary representation return iolist as a result. So, when I want to pass that result to the KafkaEx I must to convert it to a flat contiguous binary. This causes the additional usage of memory (in worst case duplicating), and guess CPU. Also if we use snappy:compress internally it is also accepts an iolist and returns flat binary, so the former binary is not need to produce.

P.S. If I'll have a time I may be perform some benchmarks and share with you.

joshuawscott:
I used the following benchmark:
```
$ cat bench/create_request_bench.exs
defmodule CreateRequestBench do
  use Benchfella

  bench "create request" do
    KafkaEx.Protocol.Produce.create_request(1, "foo", %KafkaEx.Protocol.Produce.Request{
      topic: "food", partition: 0, required_acks: 1, timeout: 10, compression: :none, messages: [
        %KafkaEx.Protocol.Produce.Message{key: "", value: "hey"}
      ]
    })
  end
end
```

Current master: 
```
$ mix bench
Settings:
  duration:      1.0 s

## CreateRequestBench
[07:06:00] 1/1: create request

Finished in 1.77 seconds

## CreateRequestBench
benchmark name  iterations   average time
create request     1000000   1.58 µs/op
```

This branch:
```
$ mix bench
Settings:
  duration:      1.0 s

## CreateRequestBench
[07:02:13] 1/1: create request

Finished in 8.4 seconds

## CreateRequestBench
benchmark name  iterations   average time
create request    10000000   0.75 µs/op
```

joshuawscott:
I should add that with a 64000 byte message the difference is even greater:
master: 22.71µs
use_iolists: 7.01µs

test/protocol/produce_test.exs
```diff
@@ -10,7 +10,7 @@ defmodule KafkaEx.Protocol.Produce.Test do
       ]
     })
 
-    assert expected_request == request
+    assert expected_request == :erlang.iolist_to_binary(request)
```
joshuawscott: I would prefer that we just change `expected_result` in these tests to a list, since that is the expected result.

lib/kafka_ex/protocol/produce.ex
```diff
@@ -32,41 +32,45 @@ defmodule KafkaEx.Protocol.Produce do
...
-  defp create_message_set([], _compression_type), do: ""
+  defp create_message_set([], _compression_type), do: {"", 0}
   defp create_message_set([%Message{key: key, value: value}|messages], :none) do
-    message = create_message(value, key)
-    message_set = << 0 :: 64-signed >> <> << byte_size(message) :: 32-signed >> <> message
-    message_set <> create_message_set(messages, :none)
+    {message, msize} = create_message(value, key)
+    message_set = [<< 0 :: 64-signed >>, << msize :: 32-signed >>, message]
+    {message_set2, ms2size} = create_message_set(messages, :none)
+    {[message_set, message_set2], 8 + 4 + msize + ms2size}
```
joshuawscott: It would be preferable if the bare numbers (8 & 4) here and in the rest of this function, and in `create_message/3` and `bytes/1` were module attributes so that it's more clear why the message set size is being padded with them.

test/protocol/produce_test.exs
joshuawscott: The `expected_request` should be changed to a list to match the expected output of the function.

dantswain:
It's tough to benchmark this.  The speed that really matters is from the function call to the data being sent out over the wire, which includes message generation and handling by the socket io.  I believe it's generally a bit more efficient to use iolists internally, we just have to be careful about making claims as to _how much_ faster.

@vjache I think this is a nice optimization.  I agree with @joshuawscott 's review, and please do try to appease ebert as much as possible (e.g., line lengths).  I know the code isn't 100% clean already but I'm trying to chip away at it.  Thanks!

bjhaid:
Thanks @joshuawscott I am convinced :)

@vjache thanks for sending in the PR

bjhaid:
The type of `topic`, here:

https://github.com/kafkaex/kafka_ex/blob/b0fef3290e1ece83bb7b997d6217557065a25896/lib/kafka_ex/protocol/produce.ex#L16

and here:

https://github.com/kafkaex/kafka_ex/blob/b0fef3290e1ece83bb7b997d6217557065a25896/lib/kafka_ex.ex#L297

needs to change to `iolist`

We need to make this change globally for sending requests and not just in the `produce` path.

dantswain:
@vjache Do you have any time to address the PR feedback?  If not, do you mind if one of us cleans this up before we merge?

joshuawscott:
Going to merge this into another branch and then fix the one test that isn't working and make a new PR.

Issue #233, opened 2017-09-22
When a connection timeout occured, fetch/2 couldn't handle the exception


tzzzoz:
Stacktrace
```
10:06:25.292 [error] Receiving data from broker "xxx.xxx.xxx.xxx":19093 failed with :timeout

10:06:25.292 [debug] Shutting down worker #PID<0.1605.0>

10:06:25.297 [error] GenServer #PID<0.1581.0> terminating
** (FunctionClauseError) no function clause matching in KafkaEx.Protocol.Fetch.parse_response/1
    (kafka_ex) lib/kafka_ex/protocol/fetch.ex:55: KafkaEx.Protocol.Fetch.parse_response(nil)
    (kafka_ex) lib/kafka_ex/server_0_p_8_p_2.ex:170: KafkaEx.Server0P8P2.fetch/2
    (kafka_ex) lib/kafka_ex/server_0_p_8_p_2.ex:77: KafkaEx.Server0P8P2.kafka_server_fetch/2
    (stdlib) gen_server.erl:615: :gen_server.try_handle_call/4
    (stdlib) gen_server.erl:647: :gen_server.handle_msg/5
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Last message: {:fetch, %KafkaEx.Protocol.Fetch.Request{auto_commit: false, client_id: nil, correlation_id: nil, max_bytes: 1000000, min_bytes: 1, offset: 2, partition: 8, topic: "raw_token_update_requests", wait_time: 10}}
```
A timeout occured while connecting to kafka brokers, `fetch/2` could not handle the `nil` response.

dantswain:
Thanks, @tzzzoz .  I've seen this a few times as well.  It is definitely a problem, and this extra context helps!

MiLk:
We always encounter this issue.
Here is the logs:
```
[error] GenServer #PID<0.594.0> terminating
** (FunctionClauseError) no function clause matching in KafkaEx.Protocol.Fetch.parse_response/1
    (kafka_ex) lib/kafka_ex/protocol/fetch.ex:55: KafkaEx.Protocol.Fetch.parse_response(nil)
    (kafka_ex) lib/kafka_ex/server_0_p_8_p_2.ex:170: KafkaEx.Server0P8P2.fetch/2
    (kafka_ex) lib/kafka_ex/server_0_p_8_p_2.ex:77: KafkaEx.Server0P8P2.kafka_server_fetch/2
[...]

PR #235, opened 2017-09-26
Avoid parsing nil response when connection timeout occured

tzzzoz:
I have done a quick fix for the issue #233.

### DIFF
--- a/lib/kafka_ex/server_0_p_8_p_0.ex
+++ b/lib/kafka_ex/server_0_p_8_p_0.ex
@@ -85,10 +85,13 @@ defmodule KafkaEx.Server0P8P0 do
         Logger.log(:error, "Leader for topic #{fetch_request.topic} is not available")
         {:topic_not_found, state}
       _ ->
-        response = broker
-          |> NetworkClient.send_sync_request(fetch_data, config_sync_timeout())
-          |> Fetch.parse_response
-        {response, %{state | correlation_id: state.correlation_id + 1}}
+        response = NetworkClient.send_sync_request(broker, fetch_data, config_sync_timeout())
+        case response do
+          nil -> {response, state}
+          _ ->
+            response = Fetch.parse_response(response)
+            {response, %{state | correlation_id: state.correlation_id + 1}}
+        end
     end
   end
 end
--- a/lib/kafka_ex/server_0_p_8_p_2.ex
+++ b/lib/kafka_ex/server_0_p_8_p_2.ex
@@ -165,21 +165,24 @@ defmodule KafkaEx.Server0P8P2 do
         Logger.log(:error, "Leader for topic #{fetch_request.topic} is not available")
         {:topic_not_found, state}
       _ ->
-        response = broker
-          |> NetworkClient.send_sync_request(fetch_data, config_sync_timeout())
-          |> Fetch.parse_response
-        state = %{state | correlation_id: state.correlation_id + 1}
-        last_offset = response |> hd |> Map.get(:partitions) |> hd |> Map.get(:last_offset)
-        if last_offset != nil && fetch_request.auto_commit do
-          offset_commit_request = %OffsetCommit.Request{
-            topic: fetch_request.topic,
-            offset: last_offset,
-            partition: fetch_request.partition,
-            consumer_group: state.consumer_group}
-          {_, state} = offset_commit(state, offset_commit_request)
-          {response, state}
-        else
-          {response, state}
+        response = NetworkClient.send_sync_request(broker, fetch_data, config_sync_timeout())
+        case response do
+          nil -> {response, state}
+          _ ->
+            response = Fetch.parse_response(response)
+            state = %{state | correlation_id: state.correlation_id + 1}
+            last_offset = response |> hd |> Map.get(:partitions) |> hd |> Map.get(:last_offset)
+            if last_offset != nil && fetch_request.auto_commit do
+              offset_commit_request = %OffsetCommit.Request{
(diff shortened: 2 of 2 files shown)


joshuawscott:
Thanks for the PR! some comments:

Ideally we'd handle this in a way that involves reconnecting, or at least returning an appropriate error.

A test would be the best way to ensure this is handled correctly (either with a reconnect or appropriate failure message). As-is, this would return `{nil, State.t}` from that function, which really just pushes the `nil` up to somewhere else, and given the limited dialyzer coverage we have, I'm not certain that will behave correctly? (correct me if I'm wrong here)

We're trying to not introduce new credo issues, so the Ebert line-too-long complaints should be addressed before merge.

Thanks again for the PR, if you are stuck on a test case, I might try to create one, as this issue is affecting the apps I'm currently working on intermittently.

Cheers!

bjhaid:
Testing this will be somewhat difficult, ideal fix will require reworking the `NetworkClient` API, I have been on the fence with this kind of fix, but given it's being there and we have not implemented a _cleaner_ fix, I think this is fine. This PR should also document that `fetch` can return `nil` instead of the datastructure it claims to return today so the user should be prepared to handle that, also the fix should be extended to the `stream` function since it's a similar behavior, and 👍  on the ebert comment

dantswain:
I'm a little bit concerned here about swallowing the failed response.  If the `nil` is due to some error state in the system, then I would rather see us raise a more meaningful error.  It would still crash, but it would be less obtuse than a match error.

Do we know for sure what conditions causes the response to be nil?  Do we know that we won't continually get nil until something crashes elsewhere?

I have been considering trying to replace the network client with something using @fishcakez Connection behavior, though that would be a bit of an undertaking.

bjhaid:
This has been a problem forever, and we have been rejecting PRs like this without fixing the underlying issue, we are also causing folks apps to crash unduly, we should give the user a choice and not make a decision for them. The ideal fix IMO is to make `NetworkClient` return a `{:ok, binary}| {:error, :reason}` and propagate that correctly to the caller, and for API like `fetch` return the error to the user and let them deal with correctly handling the error and create a `fetch!` equivalent, and for stream log it in debug mode and ignore such failures. We should also document this behavior. However if we don't have the bandwidth to adequately fix this we should stop crashing apps for folks and making people come back time and time again with this fix or asking questions why their apps are crashing.

dantswain:
I'm open to solving the underlying problem in various ways, but I'm relatively strongly against swallowing an error condition unless we know for a fact that that is not going to just cause a crash downstream somewhere.  It's possible that this will not solve the crash at all but will instead cause a user's application to crash in an even more obtuse way.

I'm very much in favor of fixing the underlying problem.  It affects me directly.  I literally sometimes have to wake up in the middle of the night to deal with its impact on our production systems.

It would be really great if someone could track down exactly why we get a nil back from the network client, under what conditions that happens, and what kind of resolution it has (does it repeat, does the next request return OK, etc).  I think @joshuawscott has some ideas around that.

bjhaid:
I am positive the `nil` is coming from here:

https://github.com/kafkaex/kafka_ex/blob/7cc8a003456ace3145e4dd5cd0d2fec64e965459/lib/kafka_ex/network_client.ex#L47

if you check your logs, you should find the entry [here](https://github.com/kafkaex/kafka_ex/blob/7cc8a003456ace3145e4dd5cd0d2fec64e965459/lib/kafka_ex/network_client.ex#L46) in it.

joshuawscott:
reason == `:timeout` in my case, (and in #233 )
I think returning the value of `{:error, reason}` would be much better, and it doesn't seem like that would require much change at all.

bjhaid:
@joshuawscott sounds like someone just volunteered :), I agree whole-heartedly with the intent

tzzzoz:
I agree with @bjhaid and @dantswain  about reworking on the NetworkClient API. When I have the connection problem in our applications, every KafkaEx API where `NetworkClient.send_sync_request/3` is involved is affected. 
Beside the issue I mentioned in #233, I got also the errors as below during the connection timeout:
[[error] Could not connect to broker "xxx.xxx.xxx.xxx":19093 because of error {:error, :econnrefused}](https://github.com/kafkaex/kafka_ex/blob/7cc8a003456ace3145e4dd5cd0d2fec64e965459/lib/kafka_ex/network_client.ex#L14)
[[error] Unable to fetch metadata from any brokers.  Timeout is 10000.](https://github.com/kafkaex/kafka_ex/blob/7cc8a003456ace3145e4dd5cd0d2fec64e965459/lib/kafka_ex/server.ex#L359)
[[error] Receiving data from broker "xxx.xxx.xxx.xxx":19093 failed with :timeout](https://github.com/kafkaex/kafka_ex/blob/7cc8a003456ace3145e4dd5cd0d2fec64e965459/lib/kafka_ex/network_client.ex#L42)

When the timeout or other connection failure comes, I think that NetworkClient should be able to handle it, either NetworkClient does the retry by itself or NetworkClient propagates the connection error to users(at least let us have the chance to capture it and deal with it), instead of crashing :kafka_ex. The connection timeout normally will last several seconds or minutes, so Supervisor's restart strategy couldn't help any more in this case.

beedub:
any update here? also seeing this issue during `:timeout`s

joshuawscott:
@tzzzoz @beedub does #272 fix this for you?

joshuawscott:
bump @tzzzoz @beedub

tzzzoz:
@joshuawscott Yes, it works for me! Great job!

PR #238, opened 2017-09-27
Fix `GenConsumer.handle_call` default implementation

### LINKED ISSUE #237: Default implementation of GenConsumer's handle_call fails dialyzer
The callback spec says the implementation must return a `{:reply...}` tuple, but the default implementation returns an atom.  This causes dialyzer to error out if the user does not implement a callback.

dantswain:
Fixes #237

The previous implementation causes a dialyzer error if the user does not
implement a handle_call.  This version is taken from the
GenServer.handle_call default implementation.

### DIFF
--- a/lib/kafka_ex/gen_consumer.ex
+++ b/lib/kafka_ex/gen_consumer.ex
@@ -272,8 +272,8 @@ defmodule KafkaEx.GenConsumer do
         # We do this to trick Dialyzer to not complain about non-local returns.
         case :erlang.phash2(1, 1) do
           0 ->
-            raise "attempted to call GenServer #{inspect proc} but no " <>
-              "handle_call/3 clause was provided"
+            raise "attempted to call KafkaEx.GenConsumer #{inspect proc} " <>
+              "but no handle_call/3 clause was provided"
           1 -> {:reply, {:bad_call, msg}, consumer_state}
         end
       end
--- a/lib/kafka_ex/gen_consumer.ex
+++ b/lib/kafka_ex/gen_consumer.ex
@@ -262,9 +262,9 @@ defmodule KafkaEx.GenConsumer do
         {:ok, nil}
       end
 
-      def handle_call(msg, _from, _consumer_state) do
+      def handle_call(msg, _from, consumer_state) do
         # taken from the GenServer handle_call implementation
-        case Process.info(self(), :registered_name) do
+        proc = case Process.info(self(), :registered_name) do
           {_, []}   -> self()
           {_, name} -> name
         end
--- a/lib/kafka_ex/gen_consumer.ex
+++ b/lib/kafka_ex/gen_consumer.ex
@@ -151,7 +151,7 @@ defmodule KafkaEx.GenConsumer do
   ```
 
   **NOTE** If you do not implement a `c:handle_call/3` callback, any calls to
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/gen_consumer.ex
```diff
@@ -262,9 +262,20 @@ defmodule KafkaEx.GenConsumer do
...
+          {_, []}   -> self()
+          {_, name} -> name
+        end
+
+        # We do this to trick Dialyzer to not complain about non-local returns.
+        case :erlang.phash2(1, 1) do
+          0 ->
+            raise "attempted to call GenServer #{inspect proc} but no " <>
+              "handle_call/3 clause was provided"
+          1 -> {:reply, {:bad_call, msg}, consumer_state}
```
joshuawscott: consumer_state isn't a valid identifier here, wouldn't this cause a compile warning?

lib/kafka_ex/gen_consumer.ex
```diff
@@ -262,9 +262,20 @@ defmodule KafkaEx.GenConsumer do
         {:ok, nil}
       end
 
-      def handle_call(_call, _from, _consumer_state) do
-        # the user must implement this if they expect to recieve calls
-        :handle_call_not_implemented
+      def handle_call(msg, _from, _consumer_state) do
+        # taken from the GenServer handle_call implementation
+        case Process.info(self(), :registered_name) do
```
joshuawscott: I think this is missing the `proc = case ...`

lib/kafka_ex/gen_consumer.ex
dantswain: :rage1:  Added a test, pushed fixes, waiting for travis.

lib/kafka_ex/gen_consumer.ex
```diff
@@ -262,9 +262,20 @@ defmodule KafkaEx.GenConsumer do
...
+        # taken from the GenServer handle_call implementation
+        proc = case Process.info(self(), :registered_name) do
+          {_, []}   -> self()
+          {_, name} -> name
+        end
+
+        # We do this to trick Dialyzer to not complain about non-local returns.
+        case :erlang.phash2(1, 1) do
+          0 ->
+            raise "attempted to call GenServer #{inspect proc} but no " <>
```
bjhaid: `GenServer` should be `GenConsumer`

lib/kafka_ex/gen_consumer.ex
bjhaid: do we need to actually return something here, it seems this will never match at least from me running in erl, is `_ -> _ # with some comment` not adequate?

lib/kafka_ex/gen_consumer.ex
dantswain: Good catch

lib/kafka_ex/gen_consumer.ex
dantswain: It should never return this in practice, but I think it's necessary for it to return a value within the dialyzer spec so that dialyzer succeeds.  I assume they did this on purpose in the GenServer implementation.

bjhaid: Ah I see, how about `_ -> {:reply, {:bad_call, msg}, consumer_state}` so the int `1` is less confusing, it's annoying to have write dumb code to just make dialyzer happy

Issue #247, opened 2017-11-15
A broker going down or coming up causes a never-ending loop of rebalances


dantswain:
We have been having hardware issues with one of our brokers.  Whenever it goes down, our consumers (using ConsumerGroup) end up in an endless cycle of rebalancing the consumer group.  The only way to resolve this is to stop all of the consumers in the group for around 60 seconds so that the broker recognizes that the consumers are all "dead", then restart the consumers.  The same thing happens when the down broker comes back online and gets elected by the cluster as a leader for some partitions.

This may be related to #246 or may be the same underlying issue - I am not 100% certain as we've been in firefighting mode.

This may also be related to the following: https://issues.apache.org/jira/browse/KAFKA-2985

dantswain:
Status update.  I'm unable to reproduce this yet with a local cluster, but it's a big issue for us in production so I am heavily invested in fixing it.

dantswain:
cjpoll in #kafkaex slack appears to be noticing a similar problem.

> We're experiencing an issue from within the Consumer Group that _seems_ to be causing quite a bit of churn for one partition. We don't seem to be able to progress past a particular point in the stream and run into this:

```
** (MatchError) no match of right hand side value: %KafkaEx.Protocol.LeaveGroup.Response{error_code: :unknown_member_id}
    (kafka_ex) lib/kafka_ex/consumer_group/manager.ex:338: KafkaEx.ConsumerGroup.Manager.leave/1
    (kafka_ex) lib/kafka_ex/consumer_group/manager.ex:167: KafkaEx.ConsumerGroup.Manager.terminate/2
    (stdlib) gen_server.erl:629: :gen_server.try_terminate/3
    (stdlib) gen_server.erl:795: :gen_server.terminate/7
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Last message: :heartbeat
  neighbours:
=CRASH REPORT==== 16-Nov-2017::22:04:55 ===
  crasher:
    initial call: Elixir.KafkaEx.ConsumerGroup.Manager:init/1
    pid: <0.4153.0>
    registered_name: []
    exception exit: {{badmatch,
                         #{'__struct__' =>
                               'Elixir.KafkaEx.Protocol.LeaveGroup.Response',
                           error_code => unknown_member_id}},
                     [{'Elixir.KafkaEx.ConsumerGroup.Manager',leave,1,
                          [{file,"lib/kafka_ex/consumer_group/manager.ex"},
[...]

PR #249, opened 2017-11-18
Stop leaking connections on consumer group shutdown/rebalance

dantswain:
This is related to #247 but also doesn't appear to be the root cause.
This should clear up the leaking gen_tcp connections that @cjpoll
observed.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -76,7 +76,7 @@ defmodule KafkaEx do
   @doc """
   Stop a worker created with create_worker/2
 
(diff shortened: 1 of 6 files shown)


lib/kafka_ex.ex
```diff
@@ -73,6 +73,17 @@ defmodule KafkaEx do
     end
   end
 
+  @doc """
+  Stop a worker created with create_worker/2
+
+  Returns `:ok` on succes or `:error` if `worker` is not a valid worker
```
joshuawscott: nit 'success'

PR #251, opened 2017-11-21
Consumer coordinator fix

### LINKED ISSUE #247: A broker going down or coming up causes a never-ending loop of rebalances

dantswain:
I believe this should fix #247 

In addition to #248 and #249 I think there were two lingering issues: 
1) Not stopping the previous consumer supervisor on rebalance.  This accounts for the `{:error, {:already_started, _}}` errors.
2) Not gracefully handling a change in consumer group coordinator node.  

I think (1) is pretty straightforward but I would appreciate some discussion on (2).  The previous behavior could fall under the "let it fail" mantra but I think it was causing enough failures to run into restart policies.  I don't think what I've done here is overly complicated - it just detects the failure and re-acquires the group metadata then recursively tries itself again.

I also had a concern about the right scope for this to happen.  Part of me says it should happen in the server implementation because one could argue that the server implementation is intended to abstract away the notion of talking to a cluster, and this is dealing with figuring out which cluster node we need to talk to.  Part of me says that following the consumer group coordinator node is part of the behavior of the consumer group itself.

The changes to the server implementation also feel a little un-DRY, but I didn't want to do any significant refactoring until the above concerns were addressed.

If this approach seems reasonable, we may actually want to implement something similar for other server transactions - e.g., update metadata whenever we get `:not_leader_for_partition`.

I'll update #247 with instructions on how to reproduce the original problem.

@bjhaid and @joshuawscott I know it's the holiday week but I'd really appreciate your input on this.  @cjpoll this hopefully fixes our issue.

### DIFF
--- a/lib/kafka_ex/consumer_group/manager.ex
+++ b/lib/kafka_ex/consumer_group/manager.ex
@@ -278,6 +278,7 @@ defmodule KafkaEx.ConsumerGroup.Manager do
     case sync_group_response do
       %SyncGroupResponse{error_code: :no_error, assignments: assignments} ->
         new_state = state
+                    |> stop_consumer()
                     |> start_consumer(unpack_assignments(assignments))
                     |> start_heartbeat_timer()
         {:ok, new_state}
--- a/lib/kafka_ex/server_0_p_9_p_0.ex
+++ b/lib/kafka_ex/server_0_p_9_p_0.ex
@@ -76,14 +76,23 @@ defmodule KafkaEx.Server0P9P0 do
   def kafka_server_join_group(join_group_request, network_timeout, state) do
     true = consumer_group?(state)
     {broker, state} = broker_for_consumer_group_with_update(state)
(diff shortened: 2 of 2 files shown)


dantswain:
I will address the Ebert issues before final merge, pending requested discussion.

dantswain:
I'm going to merge this but holding off until Monday to do a release.

PR #252, opened 2017-11-28
Fix correlation id bug, refactor consumer group network requests

dantswain:
I introduced a minor bug in #251 - the correlation id was not being properly incremented for each request.  This is not really a big issue with KafkaEx because we don't really use the correlation id to correlate responses, but I still wanted to fix it.

Part of my fix here is to refactor the consumer group sync requests to a single function.  The behavior is the same for all of the calls:

* Convert a request struct to a wire request binary
* Do the sync network request
* Parse the response binary back into a struct
* Repeat the request if we get a "not group coordinator" error message

I would like to explore using [protocols](http://elixir-lang.github.io/getting-started/protocols.html) for this, but wanted to get the fix done first.

### DIFF
--- a/lib/kafka_ex/server_0_p_9_p_0.ex
+++ b/lib/kafka_ex/server_0_p_9_p_0.ex
@@ -73,21 +73,12 @@ defmodule KafkaEx.Server0P9P0 do
     {:ok, state}
   end
 
-  def kafka_server_join_group(join_group_request, network_timeout, state_in) do
+  def kafka_server_join_group(request, network_timeout, state_in) do
     true = consumer_group?(state_in)
 
-    request_builder = fn(state) ->
-      JoinGroup.create_request(
-        state.correlation_id,
-        @client_id,
-        join_group_request
-      )
-    end
-    response_parser = &JoinGroup.parse_response/1
-
     {response, state_out} = consumer_group_sync_request(
-      request_builder,
-      response_parser,
+      request,
+      JoinGroup,
       network_timeout,
       state_in
     )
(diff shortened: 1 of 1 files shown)


lib/kafka_ex/server_0_p_9_p_0.ex
```diff
@@ -73,96 +73,92 @@ defmodule KafkaEx.Server0P9P0 do
...
 
-  def kafka_server_join_group(join_group_request, network_timeout, state) do
-    true = consumer_group?(state)
-    {broker, state} = broker_for_consumer_group_with_update(state)
-    wire_request = JoinGroup.create_request(
-      state.correlation_id,
-      @client_id,
-      join_group_request
+  def kafka_server_join_group(request, network_timeout, state_in) do
+    true = consumer_group?(state_in)
```
joshuawscott: it seems like we could drop this `true = consumer_group?(state_in)` in each of these functions and just pattern match in `consumer_group_sync_request/4`? This would also raise a match error, rather than giving a decent failure, which we could do in the shared function.
e.g.
```
def consumer_group_sync_request(_req, _mod, _timeout, %State{consumer_group: :no_consumer_group}) do
  raise "You are trying to consumer group but you aren't consumer grouping"
end
```

dantswain: Good idea.  At first I had this in the shared function, but I wanted it to error out in the parent so that you would see that in the stacktrace.  Making it raise an error and decorating the error with info from the request struct is even better :+1:

dantswain:
I think I'm going to merge this PR (unless @bjhaid objects soon) and do a new PR with the exception, since the exception would show up in more locations than what is changed here.

bjhaid:
I have not reviewed, but I am not objecting to merging, I can look tomorrow morning, but if you don't get feedback from me you can just merge :)

PR #255, opened 2017-11-29
Raise meaningful errors on misconfigured consumer group

dantswain:
From the conversation no #252


dantswain:
@bjhaid I'd appreciate your feedback here since it's a little bit of a stylistic change.

Issue #253, opened 2017-11-28
gzip test failure on macOS


dantswain:
I've just recently started seeing this error:

```
  1) test create_request correctly encodes messages with gzip (KafkaEx.Protocol.Produce.Test)
     test/protocol/produce_test.exs:30
     Assertion with == failed
     code:  assert expected_request == request
     left:  <<0, 0, 0, 0, 0, 0, 0, 1, 0, 23, 99, 111, 109, 112, 114, 101, 115,
              115, 105, 111, 110, 95, 99, 108, 105, 101, 110, 116, 95, 116, 101,
              115, 116, 0, 1, 0, 0, 0, 10, 0, 0, 0, 1, 0, 16, 99, 111, 109, 112,
              114, ...>>
     right: <<0, 0, 0, 0, 0, 0, 0, 1, 0, 23, 99, 111, 109, 112, 114, 101, 115,
              115, 105, 111, 110, 95, 99, 108, 105, 101, 110, 116, 95, 116, 101,
              115, 116, 0, 1, 0, 0, 0, 10, 0, 0, 0, 1, 0, 16, 99, 111, 109, 112,
              114, ...>>
     stacktrace:
       test/protocol/produce_test.exs:47: (test)
```

I'm on macOS 10.13.1 with Erlang 20.1.7 from homebrew and elixir 1.5.2 from kiex.  I'm not sure if this is an Elixir/OTP version problem or a mac problem.  I just checked and we're not testing against 1.5.2/20.1 in Travis, which we should be doing.  I'll open a PR to do that.

I did modify the test a bit to get it to tell me exactly which bytes were different:

```
     [{85, 79, 90}, {86, 44, 216}, {87, 46, 47}, {88, 209, 42}, {108, 3, 19}]
```

The format of this is `[{byte number, expected byte value, got byte value}]`.  The expected and got binaries are the same length.

dantswain:
#254 tested with Elixir 1.5.2 and OTP 20.1 did not reproduce this issue, so it seems to be specific to my setup.

dantswain:
I get the same error with 1.5.1, and with 1.5.2 re-installed.  I have wiped deps and _build between attempts.

It's possible these are equivalent gzipped data, maybe?

bjhaid:
maybe print out the binary data and compare?

dantswain:
@bjhaid I did (effectively) do that.  That's where this came from:

```
[...]

PR #256, opened 2017-11-29
Fix gzip test wrt changing versions of libz

dantswain:
See #253 for discussion.  There are changes to produce.ex here but they
are only formatting and refactoring that made it easier to reason about
this problem.


lib/kafka_ex/protocol/produce.ex
```diff
@@ -53,7 +56,29 @@ defmodule KafkaEx.Protocol.Produce do
...
+    %Request{
+      topic: topic,
+      partition: partition,
+      required_acks: required_acks,
+      timeout: timeout
+    },
+    message_set
+  ) do
+    KafkaEx.Protocol.create_request(:produce, correlation_id, client_id) <>
+      << required_acks :: 16-signed, timeout :: 32-signed, 1 :: 32-signed >> <>
```
joshuawscott: weird indentation here

Issue #261, opened 2017-12-06
not able to access message timestamp


dams:
Hi,

I'm consuming messages from a topic where brokers automatically set timestamps to messages when they reach them.

I'd like to access this timestamp when I'm consuming the messages, but I can't find a way to access it.

For reference, it 's the Timestamp field in the kafka message protocol, here: https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets

Is it because kafkaex doesn't support v1 MEssage format (so kafka 0.10.x) ?
If yes, any plan to support it, or any pointer to only implement support for this new message format, for a quick hack (as this feature is critical for my usage)

Thanks!

cdegroot:
I think that https://github.com/kafkaex/kafka_ex/issues/245 covers this. Of course, as usual with Open Source, the quickest way to get things done is to implement it and submit a PR ;-)

dantswain:
Hi @dams, it does look like this is because we are using the v0 format.  I am currently working may way up to the new formats, but it may take a while because I have to do it in a way that will (hopefully) work for everyone.  If you need this urgently, you might be better off forking the repo and making the changes you need :/  Of course if you come up with a good strategy for supporting the various formats, please do submit a PR!

Regardless, this is really useful feedback because it lets us know what the community's needs are.

dams:
I have implemented a great deal of the Perl Kafka driver, and the way I did it was to implement support for the api that returns what the broker is capable of, then choose the highest supported version for each api endpoint. Then each time a request is done, the apiversion is set, and kept around when we receive the reply. Then data decoding is done consistently. 

I’ll fork and patch for my need and try to do a useful PR.

joshuawscott:
We could do that if:
1. we dropped support for 0.9 and before (0.10 is the first version that supports the ApiVersionsRequest API). This isn't an option yet, since we are still using 0.9 for at least another month or so.
2. we refactor the code to allow setting versions in the various API requests. Currently everything is hard-coded to version 0.

dams:
1. I faced the same issue in the Perl client. What I did was trying to query the ApiVersionsRequest API. If it fails, we consider version 0.9, and all api endpoint versions are set to 0.9.

[...]

Issue #260, opened 2017-12-05
Call for discussion - Does anyone use the 0.8.0 or 0.8.2 implementations?


dantswain:
We maintainers are discussing dropping the 0.8.0 and 0.8.2 client implementations.

Does anyone in the community use these implementations or is everyone using the 0.9.0 implementation?

Note that existing versions of the package on hex.pm would still support 0.8.0 and 0.8.2, we would just be removing them going forward.

Feedback is appreciated!

joshuawscott:
As long as we are going to a new major version, I'd be in favor of dropping those, especially if it will make it easier to get the protocol implementation flexible enough to match the specifications for 0.9, 0.10, 0.11, 1.0 etc.

wilbertom:
@dantswain we had some conversations about an older version of Kafka so here are my two cents.

In production environments in bigger companies, we sometimes don't control what versions of a software are used. Upgrading takes a huge organizational effort to upgrade all dependent components.

That said I think your comment on older versions of the library supporting the older versions of Kafka are a great point. If someone needs support, they can just install an older version. These shouldn't updated unless one has to take a patch for security reasons.

What would be really helpful though would be a Changelog, where one could easily see when features were implemented and older things deprecated. This would be very helpful in general and people would have an easier time picking the appropriate version. 

Personally I'm all up for deleting old code, cleaning and pruning the versions that matter most in the long run. It also makes it easier for new people interested in contributing to the project, they can read just the newer implementations and not be taunted or distracted by the older implementations.

thecodeboss:
Bit late to the party, but I will add that our infrastructure still relies on Kafka 0.8.1.1 quite a bit. There have been some efforts recently to upgrade, though as @wilbertom mentioned that sort of upgrade is difficult.

Saying that, I'm okay with dropping support for future versions - I can always fork to try and get any fixes if necessary.

IanVaughan:
It would help me a lot, as I have some changes that I've only made in 0.9 server module at present, so if it's deprecated then I can hold off thinking about updating the other server version modules.

IanVaughan:
How would this be best done? Completely remove the < 0.9 server modules? Or phase it, so initially add some form of depreciation warning if the `kafka_version` is less than `0.9`?

Issue #245, opened 2017-10-24
Support the newest message formats for Kafka versions that support them


dantswain:
This will be a fairly large undertaking.

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-TheAPIs

Each message type has one or more versions, and the versions were implemented in a particular version of the broker.  The correlation is very inconsistent (e.g., v0 always starts with the first version of the broker that implements a message, so for Kafka 0.11 we may have v0 of one message and v3 of some other message).

So far we are "lucky" in that the message formats are all backwards compatible, but there are operational consequences of this.  See #244 for an example.

jbruggem:
This is now possible, since we can ask the server for the message version it supports. See an example here: https://github.com/kafkaex/kafka_ex/pull/320/files#diff-5ed86f44a0ff14f109f61199ad39cfc1L107

dantswain:
So after brainstorming a little with @joshuawscott, I decided to write some code to pull schema definitions from https://github.com/klarna/kafka_protocol (the underlying library used by brod) and use macros to generate Elixir-idiomatic structs and serde code:  https://github.com/dantswain/kayrock

I'm still tweaking it, but I believe it should work.  I have ideas about how it could be integrated with KafkaEx, but I'm super busy lately so if anyone would like to take that on feel free to reach out to me.  Ideally this would mean we can basically get rid of the entire `KafkaEx.Protocol` stack and not have to manually integrate new api versions (aside from any API logic we want to implement on this side).  It's limited to whatever messages/versions kafka_protocol has implemented, which lags a little bit but not as much as KafkaEx has ;)

joshuawscott:
@dantswain do you think this can be closed now?

PR #264, opened 2017-12-10
Refactor and test the 0.8.0 implementation

dantswain:
My ultimate goal is to first move to Protocol implementations for the wire requests and then ultimately support the various message format changes with version changes (see #261, #245).  To make these changes more manageable, I want to do some fairly heavy refactoring of the server implementations, so that they are better factored to reason about and share more implementation at the base level.

I started with the 0.8.0 implementation because it is the simplest.  Even though we are discussing dropping support for it (see #260), I think this is not wasted effort since it allows me to start with the simplest implementation and sort out some of the basic issues.

The way that I am testing this could be controversial.  I am connecting the 0.8.0 implementation to the 0.9.0 test broker with an SSL connection.  0.8.0 does not support SSL so this required allowing the 0.8.0 implementation to attempt a connection with SSL.  I am emitting a warning message when the user attempts to do this.  The only alternative that I could think of was to actually spin up a 0.8.0 cluster in test, which would significantly increase the complexity of the test setup and the duration of the test run on travis.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -471,7 +471,18 @@ defmodule KafkaEx do
     }
   end
 
-  defp build_worker_options(worker_init) do
+  @doc """
+  Builds options to be used with workers
+
+  Merges the given options with defaults from the application env config.
+  Returns p{:error, :invalid_consumer_options}` if the consumer group
+  configuation is invalid, and `{:ok, merged_options}` otherwise.
+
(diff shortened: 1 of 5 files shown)


ashwini709:
Can we get this one merged soon, if it is ok?

ashwini709:
@dantswain We are using Kafka version 0.11.0, But I could see relevant file changes that might conflict with #235 done for #233

Issue #258, opened 2017-12-04
offset_out_of_range error when starting a consumer group


thecodeboss:
When starting a consumer group, there should be a check that the consumer group offset is larger than the earliest offset available in Kafka.

For example, currently if a consumer group offset is 0, but Kafka's earliest offset is 100, it's impossible to start up a `KafkaEx.ConsumerGroup` without first requesting changes to the consumer group offset, since 

**Proposed solution:**

When consuming messages in a `KafkaEx.GenConsumer`, if an `:offset_out_of_range` error is encountered, it should automatically update the consumer offset to the earliest offset available rather than simply crashing.

dantswain:
Thanks for the issue @thecodeboss!

If I understand correctly, I'm not sure that this is something that we should handle automatically.  If you are running a consumer group and your consumer's offsets are out of range, you probably want to know that rather than have the worker quietly start consuming from the earliest offset available.  Kafka provides tools to manage consumer group offsets in situations like this.  It's also possible to `commit` the offset manually using KafkaEx.

It could be that what we should do is provide some kind of callback mechanism so that the user can decide what to do here.

thecodeboss:
@dantswain the callback mechanism would definitely be helpful.

Right now our approach during startup is to first edit our consumer offset before doing `Supervisor.start` on the `KafkaEx.ConsumerGroup`. But we still run into cases where the consumers don't keep up and things crash. It also seems like manual and unnecessary work, since the consumer should be smart enough to not request data that doesn't exist.

In `KafkaEx.GenConsumer.consume/1`, there's a match on `:no_error`, and so it always crashes without any grace when there's an `:offset_out_of_range`.

thecodeboss:
Looking at the [kafka-python repo](http://kafka-python.readthedocs.io/en/master/apidoc/kafka.consumer.html?highlight=auto_offset_reset#kafka.consumer.KafkaConsumer), it seems the way that library handles this is to have a parameter when creating the consumer called `auto_offset_reset`. You can then specify a policy for how to handle the errors automatically.
Perhaps `KafkaEx.GenConsumer` could support a similar option.

joshuawscott:
Closing as this was fixed in #266

PR #266, opened 2017-12-14
Add auto_offset_reset option to consumer implementation

thecodeboss:
This new option allows consumers to handle offset_out_of_range errors gracefully when they occur, rather than simply crashing (see #258). There are three policies to choose from when such an error happens:
1. Move consumer offset to the earliest available
2. Move consumer offset to the latest available
3. (default) let it crash
These options mimic the Java and Python Kafka libraries for how they handle these errors.

Note: I did not add any tests to this revision, as I was unsure how to approach them. We need a Kafka queue with a retention setting where old messages get deleted so that an offset_out_of_range error gets thrown.
I considered adding a new Kafka queue in the docker setup, though I wanted to get feedback on how to approach that (or even better, if one of the maintainers could help set something up that would be very useful).
I did however test this using real production topics, and as expected it stopped our consumers from crashing all the time.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule KafkaEx.Mixfile do
   def project do
     [
       app: :kafka_ex,
-      version: "0.8.2",
+      version: "0.8.1",
       elixir: "~> 1.1",
       dialyzer: [
         plt_add_deps: :transitive,
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule KafkaEx.Mixfile do
   def project do
     [
       app: :kafka_ex,
-      version: "0.8.1",
+      version: "0.8.2",
       elixir: "~> 1.1",
       dialyzer: [
         plt_add_deps: :transitive,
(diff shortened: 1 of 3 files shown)


mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule KafkaEx.Mixfile do
   def project do
     [
       app: :kafka_ex,
-      version: "0.8.1",
+      version: "0.8.2",
```
joshuawscott: The version bump is handled when we do a release, so this part should be removed.

dantswain:
Thanks, @thecodeboss, this looks pretty good!  Just a couple minor things:

* Can you please take a look and see if the new option needs to be added to any `@type`s or `@spec`s?
*  Remove the version bump.  I'll likely be doing another release in the next couple days and I'll wrap this up into that one.  This also probably deserves a minor version bump since it adds new functionality.

thecodeboss:
Thanks for the feedback @dantswain and @joshuawscott - I undid the version bump and added the open to `@type option`. I looked pretty thoroughly, seems to be the only one.

thecodeboss:
@dantswain bump, would like to get this merged

dantswain:
@thecodeboss Sorry about the delay.  I've been really busy.

PR #272, opened 2018-02-14
Avoid parsing nil response in case of connection timeout

ashwini709:
Fixes https://github.com/kafkaex/kafka_ex/issues/233
Use the code refactor in https://github.com/kafkaex/kafka_ex/pull/264

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -486,12 +486,8 @@ defmodule KafkaEx.Server do
             |> client_request(updated_state)
             |> module.create_request
 
-            response = broker
-            |> NetworkClient.send_sync_request(
-              wire_request,
-              config_sync_timeout()
-            )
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/server.ex
```diff
@@ -486,12 +486,13 @@ defmodule KafkaEx.Server do
...
-            response = broker
-            |> NetworkClient.send_sync_request(
+            response = NetworkClient.send_sync_request(
+              broker,
               wire_request,
-              config_sync_timeout()
-            )
-            |> module.parse_response
+              config_sync_timeout())
+            response = if response != nil,
```
dantswain: Oooh it's a maybe monad.

dantswain:
I'm still hesitant to swallow this error, but I'm going to give it the benefit of the doubt.

PR #274, opened 2018-03-02
Remove compile warnings

thecodeboss:
When compiling kafka_ex, there are a bunch of annoying compile warnings that get spit out on later Elixir versions like 1.5 and 1.6.

Firstly, `char_list` has been deprecated in favour of `charlist` in many places. For typespecs, we can support both by using `[char]` instead. For `String.to_char_list`, we can use `apply(String, :to_char_list, [arg])` to stop the warning, and the method exists in all Elixir versions.

Second, `Enum.partition` is deprecated, so I tweaked to remove the warning.

I also removed some unused arguments, a missing protocol implementation for `Enumerable.slice/1`, and a typo.

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -275,7 +275,7 @@ defmodule KafkaEx.Server do
       def terminate(_, state) do
         Logger.log(:debug, "Shutting down worker #{inspect state.worker_name}")
         if state.event_pid do
-          GenEvent.stop(state.event_pid)
+          :gen_event.stop(state.event_pid)
         end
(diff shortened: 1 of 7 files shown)


lib/kafka_ex/server.ex
```diff
@@ -275,7 +275,7 @@ defmodule KafkaEx.Server do
       def terminate(_, state) do
         Logger.log(:debug, "Shutting down worker #{inspect state.worker_name}")
         if state.event_pid do
-          GenEvent.stop(state.event_pid)
```
bjhaid: This is probably dead code, there should be no use of GenEvent in Kafkaex AFAIK

joshuawscott:
Thanks again @thecodeboss

PR #275, opened 2018-03-03
NetworkClient should not send nil in case of failure

ashwini709:
As pointed out by @dantswain Swallowing `nil` in https://github.com/kafkaex/kafka_ex/pull/272 was not a good idea. It leads to further crashes in here 
`response |> hd |> Map.get(:partitions) |> hd |> Map.get(:last_offset)`

screen shot 2018-02-26 at 17 25 03

If we return the `{:error, reason}` in case of network_client's `send_sync_reuqest` failure, it leads to the genserver crashing but with a valid reason.
screen shot 2018-03-03 at 10 17 01

I would like to fix the issues related to https://github.com/kafkaex/kafka_ex/issues/233 as much as possible. 
@joshuawscott @dantswain  would appreciate the guidance/suggestions to fix this behaviour. Thank you.
1. Should the network_client retry the connection?
2. If the network_client continues to timeout, should the `genconsumer's consume/1` handle the result and reply with error?

While we are at it, I would like to send `{:ok, response}` in case of success if thats ok.

### DIFF
--- a/lib/kafka_ex/server_0_p_8_p_2.ex
+++ b/lib/kafka_ex/server_0_p_8_p_2.ex
@@ -100,13 +100,15 @@ defmodule KafkaEx.Server0P8P2 do
         Logger.log(:error, "Coordinator for topic #{offset_fetch.topic} is not available")
         {:topic_not_found, state}
       _ ->
-        response = NetworkClient.send_sync_request(
-          broker,
-          offset_fetch_request,
-          config_sync_timeout())
-        response = if response != nil,
-                      do: OffsetFetch.parse_response(response),
-                      else: nil
+        response = broker
+          |> NetworkClient.send_sync_request(
+               offset_fetch_request,
+               config_sync_timeout())
+          |> case do
+               {:error, reason} -> {:error, reason}
+               response -> OffsetFetch.parse_response(response)
+             end
+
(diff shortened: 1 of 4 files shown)


lib/kafka_ex/server_0_p_8_p_2.ex
```diff
@@ -186,9 +188,15 @@ defmodule KafkaEx.Server0P8P2 do
...
+
     response = broker
-      |> NetworkClient.send_sync_request(offset_commit_request_payload, config_sync_timeout())
-      |> OffsetCommit.parse_response
+      |> NetworkClient.send_sync_request(
+           offset_commit_request_payload,
+           config_sync_timeout())
+      |> case do
+           {:error, reason} -> {:error, reason}
+           response -> OffsetFetch.parse_response(response)
```
joshuawscott: I think this should be OffsetCommit?

ashwini709: Yup. My Bad. I have updated the PR. Thanks.

ashwini709:
@joshuawscott When can we get this merged?

joshuawscott:
Going to give the other maintainers a chance to weigh in and then merge tomorrow U.S. time

joshuawscott:
Thanks again for the PR @ashwini709 !

PR #277, opened 2018-03-03
Add Elixir 1.6 to travis matrix

joshuawscott:
I also trimmed down the number of builds so that we only run a single
build for each version of Elixir.

I found 2 credo issues and fixed them. I would be in favor of running
credo and dialyzer in travis, but would like a second opinion on that.

Also updated some dependencies

### DIFF
--- a/lib/kafka_ex/protocol/produce.ex
+++ b/lib/kafka_ex/protocol/produce.ex
@@ -34,7 +34,7 @@ defmodule KafkaEx.Protocol.Produce do
   def create_request(
     correlation_id,
     client_id,
-    request = %Request{compression: compression, messages: messages}
+    %Request{compression: compression, messages: messages} = request
   ) do
     message_set = create_message_set(messages, compression)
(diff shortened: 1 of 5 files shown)


lib/kafka_ex/protocol/produce.ex
```diff
@@ -34,7 +34,7 @@ defmodule KafkaEx.Protocol.Produce do
   def create_request(
     correlation_id,
     client_id,
-    request = %Request{compression: compression, messages: messages}
+    %Request{compression: compression, messages: messages} = request
```
dantswain: I personally prefer the other way - `var = %Struct{...}`... credo just looks for consistency iirc?

joshuawscott: It does look for consistency, but I think the elixir standard library always does `%Struct{} = var`.

joshuawscott: 🤷‍♂️ I don't care either way, but I did switch a bunch to the `%Struct{} = var` in #232 😱

PR #281, opened 2018-03-29
Update documentation with potential gotcha around broker configuration

4141done:
After following the examples in the documentation I lost a few hours trying to figure out why I was getting a message that I'd successfully connected to a broker but that a leader was not present.  Hopefully this note will help others save that time when trying to get started using a stock kafka `server.properties` config file.

Please let me know if this is appropriate or if there are changes necessary to better match the goals/tone of the documentation

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,7 +1,10 @@
 use Mix.Config
 
 config :kafka_ex,
-  # a list of brokers to connect to in {"HOST", port} format
+  # A list of brokers to connect to in {"HOST", port} format. If you receive :leader_not_available
+  # errors when producing messages, it may be necessary to modify "advertised.host.name" in the
+  # server.properties file.
+  # In the case below you would set "advertised.host.name=localhost"
   brokers: [
     {"localhost", 9092},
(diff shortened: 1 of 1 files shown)


bjhaid:
@4141done :heart: thanks!

dantswain:
Thanks, @4141done !  Improvements to documentation are always welcome!

I have a minor concern but since it's already been merged it's not a huge deal.  The language you added was specific to local testing and could be misleading in some edge cases.  The list of brokers is actually a seed list - kafka will reply with a list of actual connection strings on the first successful connection.  The hostname used in that reply is the 'advertised hostname'.  This is set to localhost in the default server.properties file that most people use for testing locally.  Testing with docker can require some modification of that setting due to the way docker handles hostnames (this may be outdated at this point).  In production, it should always be the ip or hostname bound to the network interface to which you want clients to connect.

The bottom line is that configuring kafka can be tricky and I want to be careful not to lead people in the wrong direction.  I think what we have here is ok, but we should be careful not to try to document something that's outside the scope of the client library.

PR #283, opened 2018-04-02
Set sane default values for using the default worker

efcasado:
This pull-request aims at making it easier to get started with `KafkaEx`.

I've noticed that some of the settings do not provide sane defaults. This pull-request introduces the following changes:

- The default worker is disabled by default
- Sets a sane default value for the `:consumer_group` option

Without these changes, `KafkaEx` yields an error at boot time.

```elixir
Erlang/OTP 20 [erts-9.1.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

=INFO REPORT==== 2-Apr-2018::14:54:55 ===
    application: logger
    exited: stopped
    type: temporary
** (Mix) Could not start application kafka_ex: KafkaEx.start(:normal, []) returned an error: :invalid_consumer_group
```

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -533,7 +533,7 @@ defmodule KafkaEx do
     max_seconds = Application.get_env(:kafka_ex, :max_seconds, 60)
     {:ok, pid}     = KafkaEx.Supervisor.start_link(Config.server_impl, max_restarts, max_seconds)
 
-    if Application.get_env(:kafka_ex, :disable_default_worker) == true do
+    if Config.disable_default_worker do
       {:ok, pid}
     else
       case KafkaEx.create_worker(Config.default_worker, []) do
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -485,7 +485,7 @@ defmodule KafkaEx do
   def build_worker_options(worker_init) do
     defaults = [
       uris: Application.get_env(:kafka_ex, :brokers),
-      consumer_group: Application.get_env(:kafka_ex, :consumer_group),
+      consumer_group: Config.consumer_group(),
       use_ssl: Config.use_ssl(),
       ssl_options: Config.ssl_options(),
     ]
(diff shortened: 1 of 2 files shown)


joshuawscott:
Thanks for the PR!

I don't see any changes that relate to the `consumer_group` option.

Also, I am unsure that disabling the default worker makes it easier to get started. For example, `KafkaEx.fetch/3` uses the default worker `:kafka_ex` unless the `worker_name` key/value pair is in the options. This change would require creating a worker and passing that worker name along, or would require setting the configuration `disable_default_worker: false`, which seems to make it more work to get started with KafkaEx.

That's a pretty significant breaking change, since many people probably use the default worker as a way to get up and running quickly. I am not opposed to making a breaking change, but we definitely would need to change some documentation to reflect the requirement to start your own worker.

dantswain:
@efcasado Thanks for wanting to help make KafkaEx easier to use!

I agree with @joshuawscott here.  If the default worker is not enabled by default, then it is not really a "default worker".  This was originally done with the intention of making it very easy to "try" KafkaEx with little setup.

It may be worth a discussion about getting rid of the default worker completely.  It was added very early on and I have always found it a little misleading, honestly.  IMO the user should manage their own processes.  This would be a pretty big change, though, and would definitely need a) discussion and b) a major version bump.

efcasado:
Hey @joshuawscott, thanks for the prompt response 😄 

The change to the `:consumer_group` option can be found here: https://github.com/kafkaex/kafka_ex/pull/283/commits/d34e6de6b502ac0f74b12f148a1e620f945178d9

Perhaps the description of the PR was not clear enough, my bad.

It looks like I misunderstood what the default worker was all about. I thought it was a dummy worker used mostly for toying around with `KafkaEx`. I'll change the default value for `:use_default_worker` to `true` (note that prior to this PR the default value is `nil`).

efcasado:
@dantswain Thanks for the clarification. It makes a whole lot of sense. Deprecating the default worker is definitely beyond the scope of this PR. I've just updated the PR to address only the issue with the default values, which was the sole intention of this PR 😄 .

joshuawscott:
👍 This looks good to me, once the merge conflict is fixed

efcasado:
Just rebased it.

joshuawscott:
Thanks again @efcasado !

PR #285, opened 2018-04-30
Adds CSV broker definitions

hassox:
When configuring brokers from an environment variable being able to parse a list of CSV style host/ports is useful.

This is a simple change that allows the broker list in configuration to be given as either a list of CSV value.

### DIFF
--- a/lib/kafka_ex/config.ex
+++ b/lib/kafka_ex/config.ex
@@ -9,6 +9,8 @@ defmodule KafkaEx.Config do
 
   require Logger
 
+  @default_port 9092
+
   @doc false
   def disable_default_worker do
     Application.get_env(:kafka_ex, :disable_default_worker, false)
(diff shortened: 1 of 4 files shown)


joshuawscott:
I was wondering where we draw the line on this, since if you have the broker info in JSON, for example, then this doesn't work either. It seems like having a single way and just having people adapt their incoming configuration to that way is better than supporting multiple ways of specifying the broker list.

dantswain:
I agree with consistency. The reason I’d consider supporting this format is that it’s fairly standard with Kafka tools. 

One thing I’m not sure about is if this is really all that much more ops-friendly?  I think some of this might be in flux, but current SOP for deployment is to deploy a release that is going to be reading from a sys.config file, so there already has to be a step that puts the config in ETF...

hassox:
@dantswain updated the config.exs file. I had to remove it from the gitignore file to get it to be part of the PR (at least I think I needed to do that).

If this doesn't end up making it in, what is the right place to put these runtime processed items? I had a hard time trying to identify the right incantation to get everything to line up and connect.

lib/kafka_ex/config.ex
```diff
@@ -9,6 +9,8 @@ defmodule KafkaEx.Config do
 
   require Logger
 
+  @default_port 9092
```
bjhaid: I don't think we should be setting a default port, we should either log in debug mode if the pattern for `host:port` is not matched, ignore it or error

dantswain:
Thanks, @hassox!

To answer your question - one way to do this would be to set up your own supervisor and user the `uris` keyword argument.

PR #288, opened 2018-05-15
Separate Heartbeat from Manager

joshuawscott:
This moves the heartbeat functionality into a separate GenServer so that
even if the worker startup takes a long time, such as over high latency
connections, the broker won't remove the consumer from the group.

This seems to solve issues we've had where we were getting
:unknown_member_id errors back.

The heartbeat notifies the manager of a need for rebalance by exiting
with a reason of `:rebalance`. The Manager starts a rebalance upon
receiving that message.

I added a number of specs to the Manager, which I can remove if we wish - I was mainly doing this because I kept mixing up which functions return `{:ok, State.t}` vs `State.t`

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,0 +1,5 @@
+# Used by "mix format"
+[
+  inputs: ["mix.exs", "{config,lib,test}/**/*.{ex,exs}"],
+  line_length: 80
+]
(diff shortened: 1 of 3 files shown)


joshuawscott:
I did do some extensive testing manually by setting up toxiproxy and inserting up to 250ms latency (500ms RTT), and killing processes and starting them up at various points. I never really figured out how to turn that into any sort of automated testing, unfortunately. my process was basically watching to try to eliminate crash messages or thrashing. (by thrashing I mean constant rebalancing because the two are out of sync with each other)

joshuawscott:
I will do another PR to axe the specs and make them consistent

PR #290, opened 2018-05-19
Iolist cleanup

joshuawscott:
Replaces #231 

This just handles the merge conflicts and I will also clean up ebert complaints

## Original description:
This change makes using iolist instead of contiguous binary when building packet being compressed and/or sent.

P.S. Unit tests & dialyzer passed successfully but I'm failed to wait for finish integration tests, they just hanging without any log messages. Anyway, I making this PR, in case someone find time to run all integration tests.


joshuawscott:
Created #291 to track making this happen to the other protocol messages.

PR #297, opened 2018-07-18
Add a "create topics" action

### LINKED ISSUE #168: Add support for CreateTopics API
https://kafka.apache.org/protocol.html

jbruggem:
TODO:
- [x] rebase and rename my commits
- [x] clean up code to make it integrate well with existing code
- [x] add missing error codes ?
- [x] check compliance with https://github.com/kafkaex/kafka_ex/blob/master/CONTRIBUTING.md

A new server version was added because [CreateTopics was added in 0.10.1](https://issues.apache.org/jira/browse/KAFKA-2945) (see also the [release notes](https://home.apache.org/~jgus/kafka-0.10.1.0-rc3/RELEASE_NOTES.html)).

Closes #168

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
     - _build
 matrix:
   include:
+    - elixir: 1.7.3
+      otp_release: 21.0
     - elixir: 1.6.2
       otp_release: 20.2
     - elixir: 1.5.3
--- a/lib/kafka_ex/config.ex
+++ b/lib/kafka_ex/config.ex
@@ -80,6 +80,7 @@ defmodule KafkaEx.Config do
 
   defp server("0.8.0"), do: KafkaEx.Server0P8P0
   defp server("0.8.2"), do: KafkaEx.Server0P8P2
+  defp server("0.10.1"), do: KafkaEx.Server0P10P1
   defp server(_), do: KafkaEx.Server0P9P0
 
   # ssl_options should be an empty list by default if use_ssl is false
--- a/lib/kafka_ex/config.ex
+++ b/lib/kafka_ex/config.ex
@@ -80,8 +80,9 @@ defmodule KafkaEx.Config do
 
   defp server("0.8.0"), do: KafkaEx.Server0P8P0
   defp server("0.8.2"), do: KafkaEx.Server0P8P2
-  defp server("0.10.1"), do: KafkaEx.Server0P10P1
-  defp server(_), do: KafkaEx.Server0P9P0
+  defp server("0.9.0"), do: KafkaEx.Server0P9P0
+  defp server(_), do: KafkaEx.Server0P10P1
+
 
   # ssl_options should be an empty list by default if use_ssl is false
   defp ssl_options(false, []), do: []
--- a/lib/kafka_ex/protocol/create_topics.ex
+++ b/lib/kafka_ex/protocol/create_topics.ex
@@ -126,7 +126,6 @@ defmodule KafkaEx.Protocol.CreateTopics do
         |> Enum.map(function)
         |> Enum.reduce(&(&1 <> &2)))
     end
-
   end
 
   @spec parse_response(binary) :: [] | Response.t
--- a/lib/kafka_ex/protocol/create_topics.ex
+++ b/lib/kafka_ex/protocol/create_topics.ex
@@ -88,7 +88,6 @@ defmodule KafkaEx.Protocol.CreateTopics do
     replica_assignments |> map_encode(&encode_replica_assignment/1)
   end
 
-  @spec encode_replica_assignment(ReplicaAssignment.t) :: binary
   defp encode_replica_assignment(replica_assignment) do
     << replica_assignment.partition :: 32-signed >> <>
     replica_assignment.replicas |> map_encode(&(<< &1 :: 32-signed >>))
--- a/lib/kafka_ex/protocol/create_topics.ex
+++ b/lib/kafka_ex/protocol/create_topics.ex
@@ -1,0 +1,146 @@
+
+defmodule KafkaEx.Protocol.CreateTopics do
+  alias KafkaEx.Protocol
+
+  @moduledoc """
+  Implementation of the Kafka CreateTopics request and response APIs
+  """
+
+  # CreateTopics Request (Version: 0) => [create_topic_requests] timeout
+  # create_topic_requests => topic num_partitions replication_factor [replica_assignment] [config_entries]
+  #   topic => STRING
+  #   num_partitions => INT32
+  #   replication_factor => INT16
+  #   replica_assignment => partition [replicas]
+  #     partition => INT32
+  #     replicas => INT32
+  #   config_entries => config_name config_value
+  #     config_name => STRING
+  #     config_value => NULLABLE_STRING
+  # timeout => INT32
+
+  defmodule ReplicaAssignment do
+    defstruct partition: nil, replicas: nil
+    @type t :: %ReplicaAssignment{ partition: integer, replicas: [integer] }
+  end
+
+  defmodule ConfigEntry do
+    defstruct config_name: nil, config_value: nil
+    @type t :: %ConfigEntry{ config_name: binary, config_value: binary }
+  end
+
+  defmodule TopicRequest do
+    defstruct topic: nil,
+              num_partitions: nil,
+              replication_factor: nil,
+              replica_assignment: nil,
+              config_entries: nil
+    @type t :: %TopicRequest{
+      topic: binary,
+      num_partitions: integer,
+      replication_factor: integer,
+      replica_assignment: [ReplicaAssignment],
+      config_entries: [ConfigEntry],
+    }
+  end
+
+  defmodule Request do
+    @moduledoc false
+    defstruct create_topic_requests: nil, timeout: nil
+    @type t :: %Request{create_topic_requests: [TopicRequest], timeout: integer}
+  end
+
+  defmodule TopicError do
+    defstruct topic_name: nil, error_code: nil
+    @type t :: %TopicError{ topic_name: binary, error_code: integer }
+  end
+
+  defmodule Response do
+    @moduledoc false
+    defstruct topic_errors: nil
+    @type t :: %Response{topic_errors: [TopicError]}
+  end
+
+  @spec create_request(integer, binary, Request.t) :: binary
+  def create_request(correlation_id, client_id, create_topics_request) do
+    Protocol.create_request(:create_topics, correlation_id, client_id) <>
+      encode_topic_requests(create_topics_request.create_topic_requests) <>
+        << create_topics_request.timeout  :: 32-signed >>
+  end
+
+  @spec encode_topic_requests([TopicRequest.t]) :: binary
+  defp encode_topic_requests(requests) do
+    requests
+    |> map_encode(&encode_topic_request/1)
+  end
+
+  @spec encode_topic_request(TopicRequest.t) :: binary
+  defp encode_topic_request(request) do
+    encode_string(request.topic) <>
+      << request.num_partitions :: 32-signed, request.replication_factor :: 16-signed >> <>
+      encode_replica_assignments(request.replica_assignment) <>
+      encode_config_entries(request.config_entries)
+  end
+
+
+  @spec encode_replica_assignments([ReplicaAssignment.t]) :: binary
+  defp encode_replica_assignments(replica_assignments) do
+    replica_assignments |> map_encode(&encode_replica_assignment/1)
+  end
+
+  @spec encode_replica_assignment(ReplicaAssignment.t) :: binary
+  defp encode_replica_assignment(replica_assignment) do
+    << replica_assignment.partition :: 32-signed >> <>
+    replica_assignment.replicas |> map_encode(&(<< &1 :: 32-signed >>))
+  end
+
+  @spec encode_config_entries([ConfigEntry.t]) :: binary
+  defp encode_config_entries(config_entries) do
+    config_entries |> map_encode(&encode_config_entry/1)
+  end
+
+  @spec encode_config_entry(ConfigEntry.t) :: binary
+  defp encode_config_entry(config_entry) do
+    encode_string(config_entry.config_name) <> encode_nullable_string(config_entry.config_value)
+  end
+
+  @spec encode_nullable_string(String.t) :: binary
+  defp encode_nullable_string(text) do
+    if text == nil do
+      << -1 :: 16-signed >>
+    else
+      encode_string(text)
+    end
+  end
+
+  @spec encode_string(String.t) :: binary
+  defp encode_string(text) do
+    << byte_size(text) :: 16-signed, text :: binary, >>
+  end
+
+  defp map_encode(elems, function) do
+    if nil == elems or 0 == length(elems) do
+      << 0 ::  32-signed >>
+    else
+      << length(elems) ::  32-signed >> <>
+      (elems
+        |> Enum.map(function)
+        |> Enum.reduce(&(&1 <> &2)))
+    end
+
+  end
+
+  @spec parse_response(binary) :: [] | Response.t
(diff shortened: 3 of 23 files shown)


joshuawscott:
@jbruggem are you ready for review on this yet? Feel free to ping me when you are.

jbruggem:
Thanks for offering!

It's actually probably relevant to get a first review. There's probably already a lot of stuff you can point out that will help push this through.

jbruggem:
I should probably rebase first, though ?

joshuawscott:
Rebasing would be ideal, yes. I can go ahead and look it over once the conflict is resolved.

Thanks!

jbruggem:
I've rebased it.

I'm not sure if I've managed to follow the spirit of the existing code correctly in all circumstances. In particular, I had to add the notion of API version in the parsing of metadata and it felt like I didn't do it cleanly enough.

thanks again for the review.

lib/kafka_ex/config.ex
```diff
@@ -80,6 +80,7 @@ defmodule KafkaEx.Config do
 
   defp server("0.8.0"), do: KafkaEx.Server0P8P0
   defp server("0.8.2"), do: KafkaEx.Server0P8P2
+  defp server("0.10.1"), do: KafkaEx.Server0P10P1
```
bjhaid: you want to move 0.9.0 up and make this the default

lib/kafka_ex/server.ex
```diff
@@ -367,9 +383,11 @@ defmodule KafkaEx.Server do
...
       end
 
-      def update_metadata(state) do
-        {correlation_id, metadata} = retrieve_metadata(state.brokers, state.correlation_id, config_sync_timeout())
-        metadata_brokers = metadata.brokers
+      def update_metadata(state), do: update_metadata(state, nil)
+
+      def update_metadata(state, api_version) do
+        {correlation_id, metadata} = retrieve_metadata_with_version(state.brokers, state.correlation_id, config_sync_timeout(), api_version)
+        metadata_brokers = metadata.brokers |> Enum.map(&(%{&1 | is_controller: &1.node_id == metadata.controller_id}))
```
bjhaid: is this change necessary for you PR can you indicate why?

jbruggem:
@bjhaid 

>  Can you confirm if 0.9.0 supports create topic if it does then you probably don't need a new server type

A new server version was added because [CreateTopics exists only since 0.10.1](https://issues.apache.org/jira/browse/KAFKA-2945) (see also the [release notes](https://home.apache.org/~jgus/kafka-0.10.1.0-rc3/RELEASE_NOTES.html)).

> add something like this [...] to all the servers not supporting create_topic,

I already did that, but maybe not the way I should have ?

lib/kafka_ex/config.ex
jbruggem: ok, will do

bjhaid:
> I already did that, but maybe not the way I should have ?

Sorry my bad, you did the right thing!

lib/kafka_ex/server.ex
jbruggem: We had to add the `api_version` argument to functions handling metadata because in the [Metadata API](https://kafka.apache.org/protocol.html#The_Messages_Metadata) response object version 0 does not provide the `controller_id` value, which is needed to create a topic.

Therefore, it was necessary to allow support for the version 1 of the Metadata API, hence the need to specify the version in the function call.

lib/kafka_ex/protocol/metadata.ex
```diff
@@ -2,6 +2,8 @@ defmodule KafkaEx.Protocol.Metadata do
   alias KafkaEx.Protocol
   import KafkaEx.Protocol.Common
 
+  @default_api_version 0
```
bjhaid: After your previous comment wondering if we should create a new metadata_v1 module? I have not thought deeply about this just throwing it out as I imagine the more versions of things we support to more clunky the constructing request and parsing of response may become...

jbruggem: I think it's a relevant question. That's indeed the kind of question that require a deep understanding of this project, which I'm not sure I have yet.

Especially since 0.10.0.0, the client is supposed to ask the broker which API versions are supported (see "[Retrieving Supported API versions](https://kafka.apache.org/protocol.html#api_versions)"). The *"broker returns its full list of supported ApiKeys and versions"*, meaning KafkaEx should use that information to decide which API message and version to use (and allow)

This means that for version >= 0.10.0.0, KafkaEx's behaviour should be quite different than the current hard-coded mapping of versions -> code.

bjhaid: I'll leave the decision on how to proceed to @joshuawscott and @dantswain both of whom have been more involved in the project more recently than I have, however I think we should have different version modules and after looking up the supported api version either reject the api-version provided by the user or pick the newest one we have a module for and use the module... last time I looked I think there are a couple of versions of the requests that will require some amount of work to implement them, we need as much help as we can get :)

jbruggem: I'm wondering if maybe this discussion should happen in another issue, so that we can move forward with this merge request.  It seems the scale is much bigger than just adding support for a new message in the protocol. I'd be happy to help, of course !

Would that be OK ?

jbruggem:
All tests pass locally, but not on the CI. I suppose there are race conditions which become visible in that context. I tried to increase the number of retries, to no avail.

lib/kafka_ex/protocol/metadata.ex
bjhaid: @dantswain @joshuawscott what do yall think?

jbruggem:
Thanks !

lib/kafka_ex/protocol/metadata.ex
```diff
@@ -113,13 +143,55 @@ defmodule KafkaEx.Protocol.Metadata do
     parse_brokers(brokers_size - 1, rest, [%Broker{node_id: node_id, host: host, port: port} | brokers])
   end
 
+  defp parse_brokers_v1(0, rest, brokers), do: {brokers, rest}
```
joshuawscott: Attaching version numbers to the functions will become cumbersome, I think. Many of the APIs have several versions, but those version might not have any differences, which might lend itself better to having one a function parameter that is the API version.

For example, Metadata Response has no differences in the `brokers` field from versions 1-6.

lib/kafka_ex/server.ex
```diff
@@ -285,7 +289,19 @@ defmodule KafkaEx.Server do
       # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
       def kafka_server_produce(produce_request, state) do
         correlation_id = state.correlation_id + 1
-        produce_request_data = Produce.create_request(correlation_id, @client_id, produce_request)
+        produce_request_data = try do
+          Produce.create_request(correlation_id, @client_id, produce_request)
+        rescue
```
joshuawscott: How are we arriving at this exception?
Should we have a base case for `Produce.create_request/3` which returns `:invalid_produce_request` or something like that?

lib/kafka_ex/server.ex
```diff
@@ -378,17 +396,27 @@ defmodule KafkaEx.Server do
 
       # credo:disable-for-next-line Credo.Check.Refactor.FunctionArity
       def retrieve_metadata(brokers, correlation_id, sync_timeout, topic \\ []), do: retrieve_metadata(brokers, correlation_id, sync_timeout, topic, @retry_count, 0)
+
+      # credo:disable-for-next-line Credo.Check.Refactor.FunctionArity
+      def retrieve_metadata_with_version(brokers, correlation_id, sync_timeout, api_version, topic \\ []), do: retrieve_metadata(brokers, correlation_id, sync_timeout, topic, @retry_count, api_version, 0)
```
joshuawscott: We should just add `api_version` as the last argument to `retrieve_metadata` and default it to `@default_api_version` (0). No need to add another function.

lib/kafka_ex/server.ex
```diff
@@ -378,17 +396,27 @@ defmodule KafkaEx.Server do
...
+
       # credo:disable-for-next-line Credo.Check.Refactor.FunctionArity
       def retrieve_metadata(_, correlation_id, _sync_timeout, topic, 0, error_code) do
         Logger.log(:error, "Metadata request for topic #{inspect topic} failed with error_code #{inspect error_code}")
         {correlation_id, %Metadata.Response{}}
       end
+
+      # credo:disable-for-next-line Credo.Check.Refactor.FunctionArity
+      def retrieve_metadata(brokers, correlation_id, sync_timeout, topic, retry, error_code) do
```
joshuawscott: this could be removed by using a default parameter as suggested above

.travis.yml
```diff
@@ -29,6 +31,7 @@ env:
   - COVERALLS=false
 before_script:
   - IP_IFACE=eth0 ./scripts/docker_up.sh
+  - sleep 60
```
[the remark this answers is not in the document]
jbruggem: I got a bit carried away trying to make the CI pass :). I'll remove it.

lib/kafka_ex/protocol/create_topics.ex
```diff
@@ -0,0 +1,145 @@
...
+  # timeout => INT32
+
+  defmodule ReplicaAssignment do
+    defstruct partition: nil, replicas: nil
+    @type t :: %ReplicaAssignment{ partition: integer, replicas: [integer] }
+  end
+
+  defmodule ConfigEntry do
+    defstruct config_name: nil, config_value: nil
+    @type t :: %ConfigEntry{ config_name: binary, config_value: binary }
```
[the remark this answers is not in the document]
jbruggem: Indeed.

lib/kafka_ex/protocol/create_topics.ex
```diff
@@ -0,0 +1,145 @@
...
+    config_entries |> map_encode(&encode_config_entry/1)
+  end
+
+  @spec encode_config_entry(ConfigEntry.t) :: binary
+  defp encode_config_entry(config_entry) do
+    encode_string(config_entry.config_name) <> encode_nullable_string(config_entry.config_value)
+  end
+
+  @spec encode_nullable_string(String.t) :: binary
+  defp encode_nullable_string(text) do
```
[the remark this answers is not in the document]
jbruggem: > nullable string could be a nil, but the spec doesn't allow that.

So I should leave it as it is ?

> This function could use pattern matching instead of an `if` as well.

I'll replace the `if` with pattern matching.

lib/kafka_ex/protocol/metadata.ex
jbruggem: @joshuawscott agreed.

lib/kafka_ex/protocol/produce.ex
```diff
@@ -82,12 +82,16 @@ defmodule KafkaEx.Protocol.Produce do
   def parse_response(unknown), do: unknown
 
   defp create_message_set([], _compression_type), do: {"", 0}
-  defp create_message_set([%Message{key: key, value: value}|messages], :none) do
+  defp create_message_set_uncompressed([%Message{key: key, value: value}|messages]) do
```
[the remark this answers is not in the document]
jbruggem: Ok, I'll reorder.

lib/kafka_ex/server.ex
jbruggem: > How are we arriving at this exception?

For context, it's part of the changes in this commit: https://github.com/kafkaex/kafka_ex/pull/297/commits/254429e8e5143f270cd0768fe886436cca197278 .

I added this commit because I encountered a bug when I used `KafkaEx.produce` with messages that were simple maps rather than `%Message{}`. When I did that, KafkaEx would hang and the process would eat up all my system's memory very quickly. I rearranged the functions in `protocol/produce.ex` so that there would be no incorrect function match and added this exception to send back a nice error rather than just failing.

> Should we have a base case for Produce.create_request/3 which returns :invalid_produce_request or something like that?

Seems better. I'm not sure how to do it though, since the function clause error happens deeper in the stack, when `create_message_set_uncompressed` is called. Any hint ?

lib/kafka_ex/protocol/metadata.ex
```diff
@@ -93,18 +97,44 @@ defmodule KafkaEx.Protocol.Metadata do
     }
   end
 
-  def create_request(correlation_id, client_id, ""), do: KafkaEx.Protocol.create_request(:metadata, correlation_id, client_id) <> << 0 :: 32-signed >>
+  def valid_api_version(v) do
```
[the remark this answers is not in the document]
jbruggem: Ok, I'll fix that

lib/kafka_ex/protocol/metadata.ex
jbruggem: Indeed. I think it's going to become cumbersome either way, to be honest :).

I've switched to using a function parameter instead. Tell me what you think !

lib/kafka_ex/server.ex
jbruggem: done.

lib/kafka_ex/server.ex
jbruggem: done.

jbruggem:
Ok, I addressed all comments the best I could.  As usual, tests pass locally but fail on Travis (and I'm not allowed to restart them).

lib/kafka_ex/protocol/create_topics.ex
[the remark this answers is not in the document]
joshuawscott: by 'spec' I meant the typespec of the function: encode_nullable_string should take `nil | String.t`

lib/kafka_ex/server.ex
joshuawscott: I don't think I'm that concerned about it, feel free to ignore this comment.

test/integration/server0_p_10_p_1_test.exs
```diff
@@ -0,0 +1,45 @@
...
+        %{config_name: "min.compaction.lag.ms", config_value: "0"}
+      ]}
+
+    resp = KafkaEx.create_topics([request], timeout: 2000)
+    # error = NONE
+    assert {0, name} == parse_create_topic_resp(resp)
+
+    resp = KafkaEx.create_topics([request], timeout: 2000)
+    # error = TOPIC_ALREADY_EXISTS
+    assert {36, name} == parse_create_topic_resp(resp)
```
joshuawscott: We should be able to add 36 to the `@error_map` in `lib/kafka_ex/protocol.ex` and then refer to this as the error code (`:topic_already_exists`) rather than `36`

lib/kafka_ex/protocol/metadata.ex
```diff
@@ -93,24 +97,69 @@ defmodule KafkaEx.Protocol.Metadata do
     }
   end
 
-  def create_request(correlation_id, client_id, ""), do: KafkaEx.Protocol.create_request(:metadata, correlation_id, client_id) <> << 0 :: 32-signed >>
+  def create_request(correlation_id, client_id, ""), do: create_request(correlation_id, client_id, "", @default_api_version)
```
joshuawscott: minor nit: these two function heads could be consolidated by using a default value for `api_version` in `create_request/4`

jbruggem: done.

test/integration/server0_p_10_p_1_test.exs
jbruggem: Good idea, thanks ! I've added a few error codes, might as well :).

jbruggem:
OK, I think this is now ready for merge !

joshuawscott:
I'm going to wait until the end of the day to let @bjhaid or @dantswain weigh in if they wish, then I'll merge.

Thanks again!

jbruggem:
> Thanks again!

My pleasure! We were happy to be able to contribute to a nice open-source project, and it was very useful for us to have this feature implemented. I'm open to the idea of further contributions of course.

jbruggem:
awesome :tada:. Thanks for all your help !

PR #304, opened 2018-08-16
modify brokers parsing to allow runtime variables

jruhland:
Added a parser for {:system, NAME} that allows brokers to be passed in at runtime, using standard elixir pattern for managing this.

### DIFF
--- a/lib/kafka_ex/config.ex
+++ b/lib/kafka_ex/config.ex
@@ -48,10 +48,9 @@ defmodule KafkaEx.Config do
       |> brokers()
   end
 
-  defp brokers(nil),
-    do: nil
-  defp brokers(list) when is_list(list),
-    do: list
+  defp brokers(nil), do: nil
+  defp brokers({:system, env_variable}), do: brokers(System.get_env(env_variable))
+  defp brokers(list) when is_list(list), do: list
   defp brokers(csv) when is_binary(csv) do
     for line <- String.split(csv, ","), into: [] do
       case line |> trim() |> String.split(":") do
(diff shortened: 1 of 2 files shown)


joshuawscott:
Actually, looking a bit more, it doesn't seem like `{:system, variable}` is really a standard. Could this be accomplished by using the change in https://github.com/kafkaex/kafka_ex/pull/303 to load the environment variable at runtime?

greenboxal:
@joshuawscott it could but, they're not mutually exclusive. it's not really a standard but a lot of Elixir libraries follow this pattern (including Ecto and Phoenix).

I would say that for 99% of people this option works better than #303 because it's less boilerplate.

dantswain:
@greenboxal FYI Ecto no longer supports this format: https://hexdocs.pm/ecto/Ecto.Repo.html#module-urls

Personally I think the "right" solution is for KafkaEx to support something like the init callback.  I think this is entirely possible now though it might not be very intuitive.

greenboxal:
@dantswain interesting... do you to what format they're migrating now? are people starting to rely o something like distillery config providers?

dantswain:
The docs have their recommended approach, which I think is a nice solution to the problem.  It's very flexible.

I do recommend using distillery if you can.  In the past we've also used distillery to generate a release and then had our configuration management system overwrite the sys.config that distillery generates.  That might not solve the problem if you need to get config from a service discovery tool, though.  The most flexible approach is what Ecto is recommending now, which basically gives you the flexibility to supply the config however you want.  I believe KafkaEx should be compatible with that sort of approach though I'm a bit overloaded at the moment so I couldn't do the legwork to tell you exactly how :(

greenboxal:
@dantswain no worries! it makes sense. and actually, with distillery 2 they have pluggable configuration providers, which means that we can get data from the env or svc discovery. definitely considering using it with docker now

anyway, if everyone is moving towards that pattern, I'm ok with it.

jruhland:
ill go head and close this

PR #316, opened 2018-10-11
Add credo to CI tests

### LINKED ISSUE #100: Add credo for linting/style checking
Credo can be found at:

https://github.com/rrrene/credo

Whoever takes this should ensure all credo warnings/errors are cleaned up

joshuawscott:
Fixes #100

### DIFF
--- a/scripts/ci_tests.sh
+++ b/scripts/ci_tests.sh
@@ -26,4 +26,4 @@ fi
 INCLUDED_TESTS="--include integration --include consumer_group --include server_0_p_10_p_1  --include server_0_p_9_p_0 --include server_0_p_8_p_0"
 
 # Retry if it doesn't work the first time
-mix "$TEST_COMMAND" $INCLUDED_TESTS || mix "$TEST_COMMAND" $INCLUDED_TESTS
+mix "$TEST_COMMAND" "$INCLUDED_TESTS" || mix "$TEST_COMMAND" "$INCLUDED_TESTS"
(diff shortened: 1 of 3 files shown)


bjhaid:
does ebert not provide the same functionality as credo?

joshuawscott:
It does, but this gives the extra layer of ensuring that credo passes, or it will fail the test suite. It's currently passing, which was our roadblock in the past to enabling it, if I recall correctly.

scripts/ci_tests.sh
```diff
@@ -6,8 +6,15 @@
 #
 # This script could be used for local testing as long as COVERALLS is not set.
 
+set -e
```
dantswain: Instead of this, do we want to run both and then have it report which things failed?

bjhaid: `set -ex` probably will be more obvious on which failed

scripts/ci_tests.sh
```diff
@@ -16,16 +23,7 @@ else
...
-#    (we should strive to remove this but it is necessary for now)
-if [ $? -eq 0 ]
-then
-  echo "First tests passed, skipping repeat"
-else
-  echo "Repeating tests"
-  mix "$TEST_COMMAND" $INCLUDED_TESTS
-fi
+# Retry if it doesn't work the first time
+mix "$TEST_COMMAND" $INCLUDED_TESTS || mix "$TEST_COMMAND" $INCLUDED_TESTS
```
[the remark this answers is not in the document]
joshuawscott: This is incorrect. Quoting this changes the meaning of the arguments.

scripts/ci_tests.sh
joshuawscott: done

Issue #318, opened 2018-10-18
Add support for ApiVersions


jbruggem:
Supporting this would allow, starting with 0.10.0.0, to not have to individually support versions of the server in the client. KafkaEx could just determine at startup which messages are supported with which version. In the config, choosing `kafka_version: "latest"` would allow support any version starting with `0.10.0.0`.

From the doc : 

> In order to work against multiple broker versions, clients need to know what versions of various APIs a broker supports. The broker exposes this information since 0.10.0.0 as described in KIP-35. Clients should use the supported API versions information to choose the highest API version supported by both client and broker. If no such version exists, an error should be reported to the user.

See: 
* Message spec: https://kafka.apache.org/protocol.html#The_Messages_ApiVersions 
* How to handle API versions: https://kafka.apache.org/protocol.html#api_versions
* https://issues.apache.org/jira/browse/KAFKA-3307
* https://cwiki.apache.org/confluence/display/KAFKA/KIP-35+-+Retrieving+protocol+version
* https://archive.apache.org/dist/kafka/0.10.0.0/RELEASE_NOTES.htmll

PR #320, opened 2018-10-22
Use api versions for new api calls (depends on #319)

jbruggem:
See #318.

This MR replaces `server_0_p_10_p_1.ex` by `server_0_10_or_later.ex` and uses api versions to handle the API calls implemented in it.

This would mean that from here on out, all messages of the API can be implemented without assuming the version of the server. All that is needed is to refer to the `api_versions` to check which is the greatest version of the message supported both by the server and the client.

Follow-up to https://github.com/kafkaex/kafka_ex/pull/319

### DIFF
--- a/lib/kafka_ex/api_versions.ex
+++ b/lib/kafka_ex/api_versions.ex
@@ -1,0 +1,25 @@
+defmodule KafkaEx.ApiVersions do
+
+  def api_versions_map(api_versions) do
+    api_versions
+      |> Enum.map(fn version -> {version.api_key, version} end)
+      |> Map.new
+  end
+
+  def find_api_version(api_versions_map, message_type, {min_implemented_version, max_implemented_version}) do
+    if api_versions_map == [:unsupported] do
+      {:ok, min_implemented_version}
+    else
+      case KafkaEx.Protocol.api_key(message_type) do
+        nil -> :unknown_message_for_client
+        api_key ->  case api_versions_map[api_key] do
+                      %{min_version: min} when min > max_implemented_version -> :no_version_supported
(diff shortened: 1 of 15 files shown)


jbruggem:
This MR is now ready for review !

lib/kafka_ex/api_versions.ex
```diff
@@ -0,0 +1,24 @@
+defmodule KafkaEx.ApiVersions do
+
+  def api_versions_map(api_versions) do
+    api_versions
+      |> Enum.map(fn version -> {version.api_key, version} end)
+      |> Map.new
```
joshuawscott: This function isn't supported by elixir 1.1

jbruggem: thanks, I'll fix that. Any idea what to replace it with ?

jbruggem: Nevermind :)

jbruggem: From the logs of the CI? it seems I fixed that problem.

gerbal: Another option

```elixir
Enum.into(api_versions, %{}, fn version ->  {version.api_key, version} end)
```

jbruggem: thanks, I keep forgetting about `Enum.into` !

jbruggem:
Hey guys ! Anybody else apart from Joshua would like to give an opinion on this ? It is rather important in the sense that it alters a bit the way the project will implement future features, following the lead of Kafka since 0.10.

joshuawscott:
@jbruggem Thanks again! I'll try to get a release together soon

PR #325, opened 2018-11-27
Fix compile warnings and credo

joshuawscott:
Some cleanup in prep for releasing 0.9.0

### DIFF
--- a/lib/kafka_ex/protocol/create_topics.ex
+++ b/lib/kafka_ex/protocol/create_topics.ex
@@ -1,8 +1,6 @@
-
 defmodule KafkaEx.Protocol.CreateTopics do
   alias KafkaEx.Protocol
   @supported_versions_range {0, 0}
-  @default_api_version 0
 
   @moduledoc """
   Implementation of the Kafka CreateTopics request and response APIs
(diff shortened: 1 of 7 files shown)


joshuawscott:
@bjhaid I ran credo locally and fixed about a dozen issues, but ebert didn't show anything fixed. I'm wondering if something is broken there.

lib/kafka_ex/protocol/create_topics.ex
```diff
@@ -138,7 +140,7 @@ defmodule KafkaEx.Protocol.CreateTopics do
   end
 
   @spec parse_response(binary, integer) :: [] | Response.t
-  def parse_response(message, api_version)
+  #def parse_response(message, api_version)
```
bjhaid: Why is this commented out?

joshuawscott: I meant to actually remove that since it wasn't needed. Removing

bjhaid:
>  but ebert didn't show anything fixed. I'm wondering if something is broken there.

I think ebert may be having issues:

image

you should be able to login to ebert since you are an admin on this project.

PR #329, opened 2018-12-09
Add default partitioner

### LINKED ISSUE #206: DefaultPartitioner
There should be a partitioner class to that behaves similar to the [Java DefaultPartitioner](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L54-L71) so that we can deterministically write to the same partition given a key. Adding in default round-robin behavior in the case of no partition or key should also be included.

barthez:
Add default partitioner that would assign partition if one isn't set for produce request. Partition number is calculated from the message key using murmur2 hashing algorithm, to align with Java implementation. Both Java and Elixir clients will partition messages by keys in the same, deterministic way.

In the case when the key is not provided, the partition is assigned randomly.

Partitioned behavior is available to allow users to reimplement partitioner in the desired way.

Fixes #206

### DIFF
--- a/lib/kafka_ex/default_partitioner.ex
+++ b/lib/kafka_ex/default_partitioner.ex
@@ -1,0 +1,78 @@
+defmodule KafkaEx.DefaultPartitioner do
+  @moduledoc """
+  Default partitioner implementation.
+
+  When message key is set and partition isn't, partition is decided based
+  on murmur2 hash of a key to provide Java implementation consistency. When
+  message key and partition is missing, partition is selected randomly.
+  When partition is provided nothing changes.
+  """
+  use KafkaEx.Partitioner
+  alias KafkaEx.Partitioner
+  alias KafkaEx.Protocol.Produce.Request, as: ProduceRequest
+  alias KafkaEx.Protocol.Metadata.Response, as: MetadataResponse
+  alias KafkaEx.Utils.Murmur, as: Murmur
+  require Logger
+
+  @spec assign_partition(
+          request :: %ProduceRequest{},
+          metadata :: %MetadataResponse{}
+        ) :: %ProduceRequest{}
+  def assign_partition(%ProduceRequest{partition: partition} = request, _)
+      when is_number(partition) do
+    request
+  end
+
+  def assign_partition(%ProduceRequest{partition: nil} = request, metadata) do
+    case Partitioner.get_key(request) do
+      {:ok, nil} ->
+        assign_partition_randomly(request, metadata)
+
+      {:ok, key} ->
+        assign_partition_with_key(request, metadata, key)
+
+      {:error, reason} ->
+        Logger.warn(
+          "#{__MODULE__}: couldn't assign partition due to #{inspect(reason)}"
+        )
+
+        assign_partition_randomly(request, metadata)
+    end
+  end
+
+  @spec assign_partition_randomly(
+          request :: %ProduceRequest{},
+          metadata :: %MetadataResponse{}
+        ) :: %ProduceRequest{}
+  defp assign_partition_randomly(
+         %ProduceRequest{topic: topic} = request,
+         metadata
+       ) do
+    partition_id =
+      case MetadataResponse.partitions_for_topic(metadata, topic) do
+        [] -> 0
+        list -> Enum.random(list)
+      end
+
+    %{request | partition: partition_id}
+  end
+
+  @spec assign_partition_with_key(
+          request :: %ProduceRequest{},
+          metadata :: %MetadataResponse{},
+          key :: binary
+        ) :: %ProduceRequest{}
+  defp assign_partition_with_key(
+         %ProduceRequest{topic: topic} = request,
+         metadata,
+         key
+       ) do
+    hash = Murmur.umurmur2(key)
+
+    partitions_count =
+      metadata |> MetadataResponse.partitions_for_topic(topic) |> length()
+
+    partition_id = rem(hash, partitions_count)
+    %{request | partition: partition_id}
+  end
+end
--- a/lib/kafka_ex/partitioner.ex
+++ b/lib/kafka_ex/partitioner.ex
@@ -1,0 +1,45 @@
+defmodule KafkaEx.Partitioner do
+  @moduledoc """
+  Behaviour definition for partitioners, that assigns partitions for requests.
+  """
+
+  alias KafkaEx.Protocol.Produce.Request, as: ProduceRequest
+  alias KafkaEx.Protocol.Metadata.Response, as: MetadataResponse
+
+  @callback assign_partition(
+              request :: %ProduceRequest{},
+              metadata :: %MetadataResponse{}
+            ) :: %ProduceRequest{}
+
+  defmacro __using__(_) do
+    quote location: :keep do
+      @behaviour KafkaEx.Partitioner
+    end
+  end
+
+  @doc """
+  Returns key for given messages
+
+  Function looks for message key in messages list of {ProduceRequest}. It may return
+  either `{:ok, nil}` if no key was found, `{:ok, key}` when key was found,
+  or `{:error, atom}` when error happens while looking for the key.
+  """
+  @spec get_key(request :: %ProduceRequest{}) ::
+          {:ok, nil | binary} | {:error, atom}
+  def get_key(%ProduceRequest{messages: messages}) when length(messages) > 0 do
+    case unique_keys(messages) do
+      [key] -> {:ok, key}
+      _ -> {:error, :inconsistent_keys}
+    end
+  end
+
+  def get_key(_) do
+    {:error, :no_messages}
+  end
+
+  defp unique_keys(messages) do
+    messages
+    |> Enum.map(&Map.get(&1, :key))
+    |> Enum.uniq()
+  end
+end
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -370,8 +370,14 @@ defmodule KafkaEx.Server do
       # KakfaEx.Server behavior default implementations
       # This needs a refactor, but for now make credo pass:
       # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
-      def kafka_server_produce(produce_request, state) do
+      def kafka_server_produce(
+            produce_request,
+            %State{metadata: metadata} = state
+          ) do
         correlation_id = state.correlation_id + 1
+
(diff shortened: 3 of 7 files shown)


lib/kafka_ex/partitioner.ex
```diff
@@ -0,0 +1,43 @@
...
+  Behaviour definition for partitioners, that assigns partitions for requests.
+  """
+
+  alias KafkaEx.Protocol.Produce.Request, as: ProduceRequest
+  alias KafkaEx.Protocol.Metadata.Response, as: MetadataResponse
+
+  @callback assign_partition(request :: %ProduceRequest{}, metadata :: %MetadataResponse{}) ::
+    %ProduceRequest{}
+
+  defmacro __using__(_) do
```
bjhaid: Any reason why this is a macro?

barthez: Isn't it a common pattern to make behaviour defining modules define `__using__` macro, so any module that would implement the behaviour could `use BehaviousModule`?

I made `__using__` macro define `get_key` function as I believe it would be a common function used by any custom partitioner implementation.

lib/kafka_ex/partitioner.ex
```diff
@@ -0,0 +1,43 @@
...
+      @spec get_key(request :: %ProduceRequest{}) ::
+        {:ok, nil} | {:ok, binary} | {:error, binary, binary | nil}
+      def get_key(%ProduceRequest{messages: messages}) when is_list(messages) and length(messages) > 0 do
+        Enum.reduce(
+          messages,
+          {:ok, nil},
+          fn
+            (%{key: key}, {:ok, nil}) when is_binary(key) -> {:ok, key}
+            (%{key: key}, {:ok, key}) -> {:ok, key}
+            (%{key: other_key}, {:ok, key}) when is_binary(other_key) -> {:error, "Inconsistent keys within messages", key}
```
joshuawscott: If a key is set in the first message, but the second message has a key of `nil`, this 3rd clause will not catch it, and will continue as if it were set to the previously detected key.

lib/kafka_ex/partitioner.ex
joshuawscott: I think we could just make `get_key/1` a function on this module, which could be imported or used elsewhere if needed. I generally like to shy away from defining functions on another module if not needed.

lib/kafka_ex/server.ex
```diff
@@ -932,6 +933,10 @@ defmodule KafkaEx.Server do
       defp config_sync_timeout(timeout \\ nil) do
         timeout || Application.get_env(:kafka_ex, :sync_timeout, @sync_timeout)
       end
+
+      defp default_partitioner() do
```
joshuawscott: I think that the extra parenthesis here will fail in credo checking, which will fail the CI tests. You can check credo locally by running `mix credo`

test/kafka_ex/utils/murmur_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule KafkaEx.Utils.MurmurTest do
+  use ExUnit.Case
+
+  alias KafkaEx.Utils.Murmur
+
+  test "murmur2" do
+    assert Murmur.murmur2("rule") == -1673595344
```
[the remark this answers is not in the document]
yoelfme: yep, I'm seeing different results for every assert to negative numbers, but for `redundancy` or `guest` the expected value is the correct but the number are positives, also why are you using murmur2 instead of murmur3?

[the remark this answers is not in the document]
barthez: Expected values are calculated with [java implementation](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L381-L419) and confirmed with [C implementation](https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp).

You are getting a different number because the JS implementation returns result casted to uint32 (what umurmur2 is doing).

I chose `murmur2` to align with Java implementation. If the same hashing algorithm will be used, it is ensured that messages with the same keys are routed to the same partition regardless of the client.

I'm going to add `umurmur2` tests as well.

lib/kafka_ex/server.ex
barthez: No, it won't. but I can fix it to align with the project style.

lib/kafka_ex/partitioner.ex
barthez: I think it is fine. If `key` is set only on first message (or only in 1 message), it is ok to assume it is a key for all messages.

Don't you agree?

barthez:
@bjhaid @joshuawscott You can take another look.

lib/kafka_ex/default_partitioner.ex
```diff
@@ -0,0 +1,56 @@
...
+        assign_partition_with_key(request, metadata, key)
+    end
+  end
+
+  @spec assign_partition_randomly(request :: %ProduceRequest{}, metadata :: %MetadataResponse{}) ::
+    %ProduceRequest{}
+  defp assign_partition_randomly(%ProduceRequest{topic: topic} = request, metadata) do
+    partition_id =
+     case MetadataResponse.partitions_for_topic(metadata, topic) do
+       [] -> 1
```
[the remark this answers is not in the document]
joshuawscott: That's correct, this should probably be 0

lib/kafka_ex/partitioner.ex
```diff
@@ -0,0 +1,43 @@
...
+  @doc """
+  Returns key for given messages
+
+  Function looks for message key in messages list of {ProduceRequest}. It may return
+  either `{:ok, nil}` if no key was found, `{:ok, key}` when key was found,
+  or `{:error, message, nil | key}` when error happens while looking for the key.
+  """
+  @spec get_key(request :: %ProduceRequest{}) ::
+    {:ok, nil} | {:ok, binary} | {:error, binary, binary | nil}
+  def get_key(%ProduceRequest{messages: messages}) when is_list(messages) and length(messages) > 0 do
```
joshuawscott: minor nit: length(messages) already checks is_list/1

lib/kafka_ex/partitioner.ex
joshuawscott: I agree; In that case, we probably don't need to check any beyond the first, so we can just look at the head of the list and get its key, rather than iterating over a potentially large list of messages.

lib/kafka_ex/utils/murmur.ex
```diff
@@ -0,0 +1,68 @@
...
+  def murmur2(key) do
+    <<seed :: signed-size(32)>> = <<@seed :: size(32) >>
+    len = byte_size(key)
+    _murmur2(key, seed ^^^ len)
+  end
+
+  @doc """
+    Calculates murmur2 hash for given binary as unsigned 32-bit integer
+  """
+  @spec murmur2(key :: binary) :: integer
```
joshuawscott: The spec here does not match up with the function name

lib/kafka_ex/partitioner.ex
joshuawscott: On second thought, I'm unsure. Do you know how the Java client would handle this? Does it check all messages in the set, or just the first to determine partition?

lib/kafka_ex/utils/murmur.ex
```diff
@@ -0,0 +1,68 @@
...
+  def umurmur2(key) do
+    key |> murmur2() |> band(0xFFFFFFFF)
+  end
+
+  defmacrop mask32(num) do
```
bjhaid: please change this and the next to either a function or inline the code

lib/kafka_ex/utils/murmur.ex
```diff
@@ -0,0 +1,68 @@
...
+  @moduledoc """
+  Utility module that provides Murmur hashing algorithm.
+  """
+
+  use Bitwise
+
+  # Arbitrary constant for murmur2 hashing
+  @m 0x5bd1e995
+
+  # Default seed to hashing, copied form Java implementation
```
bjhaid: please link to the source of this

lib/kafka_ex/utils/murmur.ex
```diff
@@ -0,0 +1,68 @@
+defmodule KafkaEx.Utils.Murmur do
+  @moduledoc """
+  Utility module that provides Murmur hashing algorithm.
+  """
+
+  use Bitwise
+
+  # Arbitrary constant for murmur2 hashing
+  @m 0x5bd1e995
```
bjhaid: if this is copied from a reference implementation, please do link to it

joshuawscott:
Please also run `mix format` on the codebase to ensure we keep it consistent

lib/kafka_ex/partitioner.ex
barthez: Java client allows to send you 1 message at the time, so there is no problem with selecting the key ;).

I decided I will make `get_key` function to return key only when all messages in the set have the same key.

barthez:
@bjhaid @joshuawscott Thanks for your comments. I believe I addressed all of them. You can take a final look.

lib/kafka_ex/utils/murmur.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  use Bitwise
+
+  # Arbitrary constant for murmur2 hashing
+  # https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp#L39-L43
+  @m 0x5BD1E995
+  @r 24
+
+  # Default seed to hashing, copied form Java implementation
+  # https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L382
```
bjhaid: please update to:

https://github.com/apache/kafka/blob/809be928f1ae004e11d65c307ea322bef126c834/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L382

so it references the sha of the tag, if someone force pushes over the tag the line might change but it does not change with the sha

yoelfme:
wow, that's nice, I was working in a mini lib to do that for my own, but now I can use these changes, thanks  @barthez good work, when do you think that you are going to do a new release?,  but again thanks!

barthez:
@bjhaid Can we have this merged & released?

bjhaid:
> @bjhaid Can we have this merged & released?

@barthez I have not released in a while, so I'll defer doing the release to @joshuawscott, I'll merge though

PR #341, opened 2019-03-14
Added note on offset storage

gorenje:
Unfortunately it took me a while to discover the KafkaEx offsets in Zookeeper, all other services use Kafka as offset storage. 

Could please add a note in the README for future users :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,8 @@ iex> KafkaEx.metadata(topic: "foo")
 
 ### Retrieve offset from a particular time
 
+**NOTE**: KafkaEx stores offsets in Zookeeper and not in Kafka. See [#245](https://github.com/kafkaex/kafka_ex/issues/245) for more details. Also relevant: [#244](https://github.com/kafkaex/kafka_ex/issues/244), [#293](https://github.com/kafkaex/kafka_ex/issues/293), [#311](https://github.com/kafkaex/kafka_ex/issues/311) 
+
 Kafka will get the starting offset of the log segment that is created no later than the given timestamp. Due to this, and since the offset request is served only at segment granularity, the offset fetch request returns less accurate results for larger segment sizes.
 
 ```elixir


bjhaid:
Thanks for the PR, kafkaex does not directly store the offsets in zookeeper, it sends it to the broker and the broker makes the decision.

gorenje:
I read [this answer](https://stackoverflow.com/a/41150833) on StackOverflow, it does seem that it depends on what the client sends (i.e. api version 0 or 1). Also this Python code [snippet](https://github.com/dpkp/kafka-python/blob/703f06590be2daa7e4592b3d82df6d719a6829bb/kafka/protocol/commit.py#L44-L74)) seems to suggest the same. Kafka tries to be backward compatiable and hence allows older versions of the API to talk to newer servers.

As I understand, KafkaEx sends [version 0](https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/protocol.ex#L23) of the [OffsetCommitRequest](https://github.com/kafkaex/kafka_ex/blob/d9728aeebb835d8a4faa2ab557d6030219c1ea23/lib/kafka_ex/protocol/offset_commit.ex#L31) (offset commit request uses the default api version 0). 

I also experienced this running Kafka 0.11.0.3 with Node and Elixir consumers. The Node consumers wrote their offsets to Kafka and Elixir consumers wrote their offsets to Zookeeper.

bjhaid:
I think implementing the api version that writes offsets to kafka will be desirable, while maintaining backwards compatibility... this PR makes it seem kafkaex writes the offset directly to zookeeper but it does not.

If you are willing to implement support for api version 1, we'll accept the PR

gorenje:
True, what I wrote does imply that KafkaEx is writing to Zookeeper. But I think there should still be a note to point out that KafkaEx is using version 0 of the API, which in turn, means that offsets are written to Zookeeper (at least for Kafka versions 0.9 to 0.11 inclusive, higher version I don't know).

I'll try to get KafkaEx to use version 1 for at least the offset commit message. I'm not an Elixir expert but I'll have a look :)

joshuawscott:
This would definitely need to be configurable, since anyone using the current zookeeper offsets would need to migrate as per https://svn.apache.org/repos/asf/kafka/site/083/implementation.html#offsetmigration

joshuawscott:
@gorenje I think this is ok if the readme entry states that it uses the API version 0, which directs kafka to store the offsets in zookeeper.

gorenje:
I had a quick look and came up with [this approach](https://github.com/kafkaex/kafka_ex/compare/master...gorenje:offset_commit_request_version_1). Basically I added an api_version field to the Request structs and then generate either a version 1 or a version 0 protocol string in create_request. I did this for OffsetCommit and OffsetFetch. Using it looks something like this:

```elixir
iex> req = %KafkaEx.Protocol.OffsetCommit.V1.Request{ consumer_group: "kafka_ex", offset: 23, topic: "test0p8p0", partition: 0 }
%KafkaEx.Protocol.OffsetCommit.V1.Request{
  api_version: 1,
  consumer_group: "kafka_ex",
  generation_id: -1,
  member_id: "",
  metadata: "",
  offset: 23,
  partition: 0,
  timestamp: -1,
  topic: "test0p8p0"
}

iex> KafkaEx.offset_commit(:kafka_ex, req)
[%KafkaEx.Protocol.OffsetCommit.Response{partitions: [0], topic: "test0p8p0"}]

iex> fetreq = %KafkaEx.Protocol.OffsetFetch.V1.Request{ topic: "test0p8p0", partition: 0 }
%KafkaEx.Protocol.OffsetFetch.V1.Request{
  api_version: 1,
  consumer_group: nil,
  partition: 0,
  topic: "test0p8p0"
}

iex> KafkaEx.offset_fetch(:kafka_ex, fetreq)
[
  %KafkaEx.Protocol.OffsetFetch.Response{
    partitions: [
      %{error_code: :no_error, metadata: "", offset: 23, partition: 0}
    ],
    topic: "test0p8p0"
  }
]
```

I don't know whether this is the best approach but it seems to work ...

bjhaid:
nice! I think you should default to 0, and allow the users to override. and please open a PR with your change.

thanks!

gorenje:
Default remains 0, auto commit will continue using that.

I should clarify how I use KafkaEx, i.e. this approach would cover my usecase but not usecases that rely on auto-commit! Basically what I do is switch off auto commit and send an explicit commit request. I.e. deep down in my code, I've got this:

```elixir
req = %KafkaEx.Protocol.OffsetCommit.Request{
  consumer_group: consumer_group,
  topic:          topic,
  partition:      0,
  offset:         offset
}
KafkaEx.offset_commit(:kafka_ex, req)
```
So all I need to do is change the request to `%KafkaEx.Protocol.OffsetCommit.V1.Request{` and I'm done!

I would do a PR once I've tested it more and also found a why so that default behaviour can be changed (there is no way to do that at the moment). Also it needs to be refactored :)

gorenje:
Also an assumption that I make is that the response is the same for version 0 and 1 (which it is for OffsetCommit and OffsetFetch). I haven't added any code for supporting responses that differ between versions, that also needs to be dealt with. (Unfortunately Kafka doesn't include the API version in it's responses, if it did, then this would be trivial thing to solve.)

dantswain:
I'm closing this as I believe the new client makes this distinction more clear and now allows users to use the versions of the API that store offsets in Kafka.

PR #342, opened 2019-03-19
Add OffsetCommit and OffsetFetch API Version 1

gorenje:
This adds code to support for OffsetCommit and OffsetFetch API Version 1. API version 1 is the Kafka backed storage for offsets, while version 0 uses Zookeeper.

What this does is add the support but does not change the default behaviour of KafkaEx in using version 0. To change the default behaviour, I suspect that offset handling also needs to change. For
example, I suspect that [offset + 1](https://github.com/kafkaex/kafka_ex/blob/3c1eaf8b7c7925fc790d7bcea93de7b949638348/lib/kafka_ex.ex#L516) would not need to be done when using kafka-backed offset storage. But I'm not 100% sure.

This approach tries to be keep code changes to a minimum, I'm sure there are other ways to achieve this. Also I don't know how this would affect #245?


gorenje:
Travis fails on 1.6.1 and 1.1.1 on something I suspect is a timing/sync issue with Kafka. Is there a good way to solve this (other than adding a sleep ;))?

bjhaid:
I have kicked them again

gorenje:
I was thinking about how this might be completely implemented, that is, having KafkaEx do version 1 commits and requests seemlessly.

This would definitely be something that globally configured and at runtime, it's not something that changes dynamically. So I got as far as having a `use_v1_commit` configuration option but then got stuck :) 

I came across this in the config.exs:

```elixir
  defp server("0.8.0"), do: KafkaEx.Server0P8P0
  defp server("0.8.2"), do: KafkaEx.Server0P8P2
  defp server("0.9.0"), do: KafkaEx.Server0P9P0
  defp server(_), do: KafkaEx.Server0P10AndLater
```

if there was another "server" for v1 versus v0, then that might be the cleanest approach. But because of all the `defdelegate`, it would be messy to have an `Server0P8P2v1` so maybe not a good approach.

From what I understand from the code, there only are 3 functions that need touching for this: [KafkaEx.stream](https://github.com/kafkaex/kafka_ex/blob/3c1eaf8b7c7925fc790d7bcea93de7b949638348/lib/kafka_ex.ex#L505), [KafkaEx.current_offset](https://github.com/kafkaex/kafka_ex/blob/3c1eaf8b7c7925fc790d7bcea93de7b949638348/lib/kafka_ex.ex#L574) and [Server0P8P2.fetch](https://github.com/kafkaex/kafka_ex/blob/3c1eaf8b7c7925fc790d7bcea93de7b949638348/lib/kafka_ex/server_0_p_8_p_2.ex#L268). So something like a `defdelegate` might also cut it or clever `alias` ... either way, I'm still thinking about a good, clean way of implementing it fully....

gorenje:
I've gone and made the [changes (rough draft)](https://github.com/gorenje/kafka_ex/compare/offset_commit_request_version_1...gorenje:use_v1_commit) of how this PR could be used within KafkaEx. The changes are based off the code in this PR, the diff doesn't show the changes made by this PR.

Not sure if what I did the best way to do this, feedback would be much appreciated!

FWIW, I've tried this out on my setup and the offsets are indeed now written to Kafka :)

joshuawscott:
@gorenje sorry about the delay in looking at this, I just got back from a vacation, and I'm trying to catch up. I'll take a look within the next day.

gorenje:
@joshuawscott that would be great and no worries about timing, I've got my fork and I can use that so that's fine :)

I guess what is important is whether this fits into the bigger scheme of things and whether this a solution that can be used moving forward, in particular, when it comes to #245.

lib/kafka_ex/protocol/offset_commit.ex
```diff
@@ -29,6 +30,28 @@ defmodule KafkaEx.Protocol.OffsetCommit do
 
   @spec create_request(integer, binary, Request.t()) :: binary
   def create_request(correlation_id, client_id, offset_commit_request) do
+    create_request_by_version(correlation_id, client_id,
+      offset_commit_request, offset_commit_request.api_version)
+  end
+
+  defp create_request_by_version(
```
joshuawscott: Could we perhaps just collapse this down into a single function with an optional `api_version_number` parameter?

gorenje: The main reason is that I didn't want to change the signature of the `create_request` method. I.e. I wanted to keep the code changes to a minimum and not also make changes to server.ex and server_p0p8p2.ex.

Another thought was consistency amongst all `create_request` signatures but I just noticed that the create_request signatures do vary: [metadata has api_version](https://github.com/kafkaex/kafka_ex/blob/455fb652cfbf3554c885522a60329bd424c9aa60/lib/kafka_ex/protocol/metadata.ex#L148-L153), [fetch just takes fetch_request](https://github.com/kafkaex/kafka_ex/blob/455fb652cfbf3554c885522a60329bd424c9aa60/lib/kafka_ex/protocol/fetch.ex#L60) - taking correlation_id and client_id from the request, [heartbeat has correlation_id, client_id and request](https://github.com/kafkaex/kafka_ex/blob/455fb652cfbf3554c885522a60329bd424c9aa60/lib/kafka_ex/protocol/heartbeat.ex#L26) and [create_topics has api_version included](https://github.com/kafkaex/kafka_ex/blob/455fb652cfbf3554c885522a60329bd424c9aa60/lib/kafka_ex/protocol/create_topics.ex#L81-L86).

One minor point is that the way this is done (i.e. including api_version in the Request struct) would mean that the call becomes something like

    create_request(correlation_id, client_id, request, request.api_version)

so the api_version comes from the request, which is kinda doubling up when the request is already being passed in.

For me, it all starts with where to store the api_version. If it's stored directly in the struct defining the request, then passing it in explicitly, when the request is already being passed in, makes no sense. There are other ways to distinguish the various api_versions' from one another. For me, how the api_version is represented will define, in part, the signature for `create_request` across the board.

Sorry this got a bit ranty :)

lib/kafka_ex/protocol/offset_commit.ex
```diff
@@ -11,7 +11,8 @@ defmodule KafkaEx.Protocol.OffsetCommit do
               topic: nil,
               partition: nil,
               offset: nil,
-              metadata: ""
+              metadata: "",
+              api_version: 0
 
     @type t :: %Request{
```
joshuawscott: Does this type need to be updated?

lib/kafka_ex/protocol/offset_commit.ex
joshuawscott: I think we can probably leave api_version out of the Request struct, since ideally we would actually check with the server to know what versions it supports before sending a request (for example, if the version isn't provided. We should also probably set a constraint on which versions we supports (as-is, if version 2 came out, it would send version 0 right now, which would probably not be desirable)

lib/kafka_ex/protocol/offset_commit.ex
gorenje: Without it, the api_version can't be used consistently, i.e. [this line](https://github.com/kafkaex/kafka_ex/pull/342/files#diff-23354f817459dba990c9cab84bd1ab76R34) would need to check the type of the offset_commit_request

lib/kafka_ex/protocol/offset_commit.ex
gorenje: Checking what the server/Kafka can and can't do doesn't cover all use-cases. I.e. if Kafka can store the consumer offsets in either Kafka or Zookeeper, then KafkaEx would always store in Kafka (since this represents the higher api version) even though the user might want to support legacy software that stores to Zookeeper.

That's why other Kafka libraries (i.e. python and nodejs) tend not to do so much magic and instead make the user decide which api versions they would like to use.

Something else that I noticed was that KafkaEx stores the offset of the last-read message not the offset of the next-to-read message. This then means that the consumer offsets are always one behind the offset of the topic, making it appear that the consumer still has a message to consumer. When requesting new messages, KafkaEx then adds one to the current consumer offset and all is well. Other libraries set the offset to the next-to-read offset and don't add one to the offset when requesting more messages. This then avoids the appearance that the consumer still has a message to consumer.

joshuawscott: When I read the python docs at https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html it says:

* *api_version (tuple) –*
Specify which Kafka API version to use. If set to None, the client will attempt to infer the broker version by probing various APIs. Different versions enable different functionality.
*Examples*
`(0, 9)` enables full group coordination features with automatic
partition assignment and rebalancing,
`(0, 8, 2)` enables kafka-storage offset commits with manual
partition assignment only,
`(0, 8, 1)` enables zookeeper-storage offset commits with manual
partition assignment only,
`(0, 8, 0)` enables basic functionality but requires manual
partition assignment and offset management.

We do want to provide a way to select where to store the offsets, since migration will be needed to get from zookeeper to kafka, but I think that could be a configuration setting that's more user friendly than having to know API version 0 vs 1 for a specific protocol message.

gorenje: Of course, having one central configuration to enable whatever use-case the user would needs is ideal. However you'll need to differentiate between protocol versions because requests and response vary between versions. 

What I meant when I mentioned python was something along the lines of the following:

```python
cli = KafkaClient(bootstrap_servers=broker_hosts, client_id='client_id')
request = ListGroupsRequest[1]()
for broker_id in broker_ids:
    future = cli.send(broker_id, request)
    cli.poll(future=future)
    ...
```

where the 1 in the line `ListGroupsRequest[1]()` defines which api version the user wants. If I wanted to use api version 0, then all I have to do is replace the 1 with a 0, or a 2 for that matter for version 2. The use-case I'm dealing with here is introspection of the Kafka and what offset values I have for various consumers.

If I was only doing consumption and production of messages, then I won't be interested in version numbers, I would only be interested in where the offset was stored and which partition has the data I need. And for that, one central configuration would be perfect.

At the moment, I'm only doing consumption with KafkaEx and for that, this change is sufficient since I only want to write my offsets to Kafka and not Zookeeper. If I wanted to do the same introspection that I'm doing with Python with KafkaEx, then that won't be as simple.

joshuawscott: So, we've had a long discussion here, but I think my feedback is that this function could be simplified by having a default parameter, without having to introduce a second function that takes a version number.
I also think we need to keep the version number out of the protocol struct, since we are already putting it as an argument in the function, and the other places that use a protocol api version do not add it to the protocol struct.

dantswain:
Hi @gorenje, thanks for this PR!  Do you mind if @joshuawscott or I fork your branch and finish polishing this up?

gorenje:
@dantswain go for it!

zapient:
What will the process for switching from ZK to Broker storage without losing committed offsets look like?

lib/kafka_ex/protocol/offset_commit.ex
jbruggem: (disclaimer: I might be understanding this incorrectly)

It seems to me that the API version should not be part of the request. Instead, we should determine what versions are supported by the server and use the highest.

See here for an example: https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/server_0_p_10_and_later.ex#L184

My proposition means that:
- the version would be an additional argument of `KafkaEx.Protocol.OffsetCommit.create_request/3`
- in `server_0_p_8_*` and `server_0_p_9_*`, the version would be fixed
- in `server_0_p_10_and_later`, we would get the version from the server

Though this might be what's going to happen here and I just didn't understand correctly :).

joshuawscott:
Based on the work that's gone into updating the APIs to support versioning by @dantswain, I'm going to close this.

gorenje:
:+1:

PR #343, opened 2019-03-19
Elixir 1.8.1 warnings could be removed

gorenje:
I get some warnings using 1.8.1:

    warning: invalid type annotation. When using the | operator to represent the union of types, make sure to wrap type annotations in parentheses: nil | number | opts :: Keyword.t()
      lib/kafka_ex/server.ex:232

and

    warning: System.cwd/0 is deprecated. Use File.cwd/0 instead
      config/config.exs:60

    warning: System.cwd/0 is deprecated. Use File.cwd/0 instead
      config/config.exs:61

    warning: System.cwd/0 is deprecated. Use File.cwd/0 instead
      config/config.exs:62

this would remove them but dunno what it does on other Elixir versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
     - _build
 matrix:
   include:
+    - elixir: 1.8.1
+      otp_release: 21.1
     - elixir: 1.7.4
       otp_release: 21.1
     - elixir: 1.6.6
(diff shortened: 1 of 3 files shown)


bjhaid:
can you add 1.8.1 to the mix here:

https://github.com/kafkaex/kafka_ex/blob/3c1eaf8b7c7925fc790d7bcea93de7b949638348/.travis.yml#L10-L25

❤️

gorenje:
Thank you for a great package :)

PR #344, opened 2019-03-27
Allow manual offset tracking for GenConsumer

netDalek:
Allow to specify exact commit offset during startup and allow disable offset commit in handle_message_set of GenConsumer

### DIFF
--- a/lib/kafka_ex/gen_consumer.ex
+++ b/lib/kafka_ex/gen_consumer.ex
@@ -829,23 +829,8 @@ defmodule KafkaEx.GenConsumer do
            partition: partition
          } = state
        ) do
-    request = %OffsetFetchRequest{
-      consumer_group: group,
-      topic: topic,
-      partition: partition
-    }
-
-    [
-      %OffsetFetchResponse{
-        topic: ^topic,
-        partitions: [
-          %{partition: ^partition, error_code: error_code, offset: offset}
-        ]
-      }
-    ] = KafkaEx.offset_fetch(worker_name, request)
-
-    case error_code do
-      :no_error ->
+    case Keyword.fetch(state.fetch_options, :offset) do
+      {:ok, offset} ->
         %State{
           state
           | current_offset: offset,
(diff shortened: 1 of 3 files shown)


joshuawscott:
Thanks for the PR! I have questions on this:
1. How does this avoid re-processing messages after returning from the handle_message_set function? If the commits aren't done at that point, it's going to fetch the messages from the last commit point, which could be duplicates. Put another way: how does this differ from the current method of being able to commit from the consumer, but still ensuring a synchronous commit is done when the function returns?
2. Can we add documentation spelling out the differences in the strategies to answer question 1?
3. Can we add test cases to ensure that we aren't re-processing messages?

Thanks!

netDalek:
Thanks for quick response.
1. Fetch use `current_offset` value which changed in GenConsumer.handle_message_set function. Or you mean about the case with whole GenConsumer restart? I think no_commit can be used mostly in two cases. First is when we are not interested in storing offset in kafka. It may be some kind of compacted topic that we always read from the beginning and cache in local ets (the case I'm trying now). Or we store processing result in some RDBMS and decide to store commit in RDBMS too in the same transaction. On the other way we can store processing result in rotating files, and to be more transactional we want to commit only after closing the file. This way handle_message_set would return sync_commit only after closing file and return no_commit in other cases. I think this is some kind of two phase commit (2PC). 

I will try to do 2 and 3

netDalek:
I've add simple test for no_commit strategy. Could you take a look please. Is it what you are talking about?

joshuawscott:
Test looks good, but I have some concerns about the goal of the functionality as a whole:

I think the case I was concerned with is not only a shutdown, but also a rebalance, where another worker on another machine could potentially begin consuming from the last committed offset. In that case, the new consumer wouldn't start consuming from the beginning of the topic, unless the old consumer never committed an offset. While this is a valid use case, GenConsumer is intended to coordinate work between even non-clustered servers, and this is explicitly deciding not to coordinate the point in the offsets.

For the first use case, I would recommend https://hexdocs.pm/kafka_ex/KafkaEx.html#stream/3 the stream API. This would allow storing offsets, and would start from the beginning of the topic every time if you wished.

For the second use case, since committing an offset necessarily commits all offsets before it, you couldn't be guaranteed that a commit means that work is complete unless you processed the messages strictly in order, in which case sync_commit would accomplish the same thing as no_commit.

In short, I'm a little hesitant to increase the complexity of GenConsumer unless its really needed, so if the functionality can be achieved by other means, I would favor that. I'm still open to no_commit if there are use cases that are much better using GenConsumer with no_commit.

netDalek:
Thank you for the explanation. Simplicity is very important, thats why kafka_ex looks much better than brod ). And I'm looking forward to be able to store offset commit in kafka that discussed in sibling pull request.

I haven't analyse rebalancing behaviour and streams a lot for now. And now I even think that some of my use cases should be implemented with raw fetch methods. So I'll close this PR.

Issue #298, opened 2018-07-18
How to handle "Could not connect to broker"?


IanVaughan:
If the Kafka broker is not present/found, then the `KafkaEx.ConsumerGroup` crashes, which brings down the whole app.

```
10:46:06.838 [error] Could not connect to broker "localhost":9092 because of error {:error, :econnrefused}

10:46:06.852 [error] Unable to fetch metadata from any brokers.  Timeout is 3000.
10:46:06.857 [error] CRASH REPORT Process <0.1003.0> with 0 neighbours exited with reason: #{'__exception__' => true,'__struct__' => 'Elixir.RuntimeError',message => <<"Unable to fetch metadata from any brokers.  Timeout is 3000.">>} in 'Elixir.KafkaEx.Server0P9P0':retrieve_metadata/6 line 393 in gen_server:init_it/6 line 352
10:46:06.857 [error] CRASH REPORT Process <0.1002.0> with 0 neighbours exited with reason: no match of right hand value {error,{#{'__exception__' => true,'__struct__' => 'Elixir.RuntimeError',message => <<"Unable to fetch metadata from any brokers.  Timeout is 3000.">>},[{'Elixir.KafkaEx.Server0P9P0',retrieve_metadata,6,[{file,"lib/kafka_ex/server.ex"},{line,393}]},{'Elixir.KafkaEx.Server0P9P0',kafka_server_init,1,[{file,"lib/kafka_ex/server_0_p_9_p_0.ex"},{line,67}]},{gen_server,init_it,6,[{file,"gen_server.erl"},{line,328}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,247}]}]}} in 'Elixir.KafkaEx.ConsumerGroup.Manager':init/1 line 97 in gen_server:init_it/6 line 352
[...]

PR #347, opened 2019-05-05
Attempt of better connection error handling / reconnect

ukrbublik:
Related to issues #298, #190 

When kafka connection is lost, handle relative connection errors gracefully.
Currently supervisor tree will try to recreate itself without timeout which leads to quick exhaustion of max restarts and complete shutdown of application :kafka_ex.
I added sleep (config `sleep_for_reconnect`, default 400 ms) in server's `init` at places that throws errors without connection to kafka.

Tested with `docker-compose down && docker-compose up`. 
Application :kafka_ex will not crash but contantly  (with accordingly configured `max_restarts/max_seconds/sleep_for_reconnect`, 3/1/500 for example) tries to reconnect.
Or if configured with `max_restarts/max_seconds/sleep_for_reconnect`, 20/10/400 for example will try to reconnect t Kafka in ~10s and then shutdown.

Tests pass.

### DIFF
--- a/lib/kafka_ex/consumer_group/heartbeat.ex
+++ b/lib/kafka_ex/consumer_group/heartbeat.ex
@@ -79,6 +79,11 @@ defmodule KafkaEx.ConsumerGroup.Heartbeat do
       %HeartbeatResponse{error_code: error_code} ->
         Logger.warn("Heartbeat failed, got error code #{error_code}")
         {:stop, {:shutdown, {:error, error_code}}, state}
+
+      {:error, reason} ->
+        Logger.warn("Heartbeat failed, got error reason #{inspect reason}")
+        {:stop, {:shutdown, {:error, reason}}, state}
+
     end
   end
 end
--- a/lib/kafka_ex/consumer_group/manager.ex
+++ b/lib/kafka_ex/consumer_group/manager.ex
@@ -192,15 +192,30 @@ defmodule KafkaEx.ConsumerGroup.Manager do
     {:noreply, state}
   end
 
+  # If the heartbeat gets an unrecoverable error.
+  def handle_info(
+        {:EXIT, _heartbeat_timer, {:shutdown, {:error, _reason}}},
+        %State{} = state
+      ) do
+    {:noreply, state}
+  end
+
   # When terminating, inform the group coordinator that this member is leaving
   # the group so that the group can rebalance without waiting for a session
   # timeout.
-  def terminate(_reason, %State{generation_id: nil, member_id: nil}), do: :ok
+  def terminate(_reason, %State{generation_id: nil, member_id: nil} = state) do
+    Process.unlink(state.worker_name)
+    KafkaEx.stop_worker(state.worker_name)
+  end
 
   def terminate(_reason, %State{} = state) do
     {:ok, _state} = leave(state)
     Process.unlink(state.worker_name)
     KafkaEx.stop_worker(state.worker_name)
+
+    # should be at end because of race condition (stop heartbeat while it is shutting down)
+    # if race condition happens, worker will be abandoned
+    stop_heartbeat_timer(state)
   end
 
   ### Helpers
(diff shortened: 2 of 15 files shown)


lib/kafka_ex/consumer_group/heartbeat.ex
```diff
@@ -79,6 +79,11 @@ defmodule KafkaEx.ConsumerGroup.Heartbeat do
       %HeartbeatResponse{error_code: error_code} ->
         Logger.warn("Heartbeat failed, got error code #{error_code}")
         {:stop, {:shutdown, {:error, error_code}}, state}
+
+      {:error, reason} ->
```
joshuawscott: It's not possible for this value to be returned from `KafkaEx.heartbeat/2` according to dialyzer.

lib/kafka_ex/consumer_group/manager.ex
```diff
@@ -244,9 +259,14 @@ defmodule KafkaEx.ConsumerGroup.Manager do
...
     # error message
-    if join_response.error_code != :no_error do
-      raise "Error joining consumer group #{group_name}: " <>
-              "#{inspect(join_response.error_code)}"
+    case join_response do
+      %{error_code: :no_error} -> :ok
+      %{error_code: error_code} ->
+        raise "Error joining consumer group #{group_name}: " <>
+                "#{inspect(error_code)}"
+      {:error, reason} ->
```
joshuawscott: Same here - dialyzer says that join_response can never be `{:error, _}`

lib/kafka_ex/consumer_group/manager.ex
```diff
@@ -341,13 +360,19 @@ defmodule KafkaEx.ConsumerGroup.Manager do
...
-      end)
+    case leave_group_response do
+      %{error_code: :no_error} ->
+        Logger.debug(fn -> "Left consumer group #{group_name}" end)
+      %{error_code: error_code} ->
+        Logger.warn(fn ->
+          "Received error #{inspect(error_code)}, " <>
+            "consumer group manager will exit regardless."
+        end)
+      {:error, reason} ->
```
joshuawscott: Also here, Dialyzer catches this as not possible

lib/kafka_ex/consumer_group/heartbeat.ex
ukrbublik: I'm sure it's possible, I've checked this on live application.

KafkaEx.heartbeat is called here - 
https://github.com/kafkaex/kafka_ex/blob/a746114bd2a673b438bb88234eb7966d556288df/lib/kafka_ex/consumer_group/heartbeat.ex#L69
->
KafkaEx.heartbeat  https://github.com/kafkaex/kafka_ex/blob/a746114bd2a673b438bb88234eb7966d556288df/lib/kafka_ex.ex#L146
->
KafkaEx.Server.handle_call({:heartbeat... - https://github.com/kafkaex/kafka_ex/blob/a746114bd2a673b438bb88234eb7966d556288df/lib/kafka_ex/server.ex#L336
->
KafkaEx.Server0P9P0.kafka_server_heartbeat  - https://github.com/kafkaex/kafka_ex/blob/a746114bd2a673b438bb88234eb7966d556288df/lib/kafka_ex/server_0_p_9_p_0.ex#L186
->
KafkaEx.Server0P9P0.consumer_group_sync_request - https://github.com/kafkaex/kafka_ex/blob/a746114bd2a673b438bb88234eb7966d556288df/lib/kafka_ex/server_0_p_9_p_0.ex#L228

lib/kafka_ex/consumer_group/manager.ex
ukrbublik: Also checked on live application - exception can be raised from line 268

lib/kafka_ex/consumer_group/heartbeat.ex
joshuawscott: heh, that's unfortunate. I suspect we need to fix the `@spec` for KafkaEx.heartbeat then, so that dialyzer will pass.

ukrbublik:
@joshuawscott I've updated specs, Dialyzer pass.

PR #349, opened 2019-05-21
Refresh metadata after creating or deleting a topic

jbruggem:
This PR corrects what I believe is a bug in the implementation of `create_topics` and `delete_topics`, which leads to stale metadata in the GenServer's state.

Indeed, once the creation/deletion of a topic has happened, the metadata in the state of the worker is stale and does not represent the real state of the server.

Hence we should update those metadata when we do that creation or deletion.

I also took the opportunity to rename the callbacks I had added because they were not consistent with the other ones.

Edit: added some fixes to specs (small inconsistencies that were making our dialyzer fail downstream).

### DIFF
--- a/lib/kafka_ex/server_0_p_10_and_later.ex
+++ b/lib/kafka_ex/server_0_p_10_and_later.ex
@@ -95,6 +95,7 @@ defmodule KafkaEx.Server0P10AndLater do
     check_brokers_sockets!(brokers)
 
     {_,
+
     %KafkaEx.Protocol.ApiVersions.Response{
       api_versions: api_versions,
       error_code: error_code
@@ -181,7 +182,7 @@ defmodule KafkaEx.Server0P10AndLater do
     {:noreply, update_metadata(state)}
   end
 
-  def kafka_api_versions(state) do
+  def kafka_server_api_versions(state) do
     response =
       state.correlation_id
       |> ApiVersions.create_request(@client_id)
@@ -191,7 +192,7 @@ defmodule KafkaEx.Server0P10AndLater do
     {:reply, response, %{state | correlation_id: state.correlation_id + 1}}
   end
 
-  def kafka_delete_topics(topics, network_timeout, state) do
+  def kafka_server_delete_topics(topics, network_timeout, state) do
     api_version =
       case DeleteTopics.api_version(state.api_versions) do
         {:ok, api_version} ->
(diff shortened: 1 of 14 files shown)


test/integration/server0_p_10_and_later_test.exs
```diff
@@ -19,11 +20,14 @@ defmodule KafkaEx.Server0P10P1AndLater.Test do
     resp = create_topic(name, config)
     assert {:topic_already_exists, name} == parse_create_topic_resp(resp)
 
-    wait_for(fn ->
-      Enum.member?(existing_topics(), name)
-    end)
+    Enum.member?(existing_topics(), name)
```
jbruggem: here, we don't wait anymore since the metadata will be refreshed immediatly

lib/kafka_ex/server_0_p_10_and_later.ex
```diff
@@ -224,10 +224,12 @@ defmodule KafkaEx.Server0P10AndLater do
           {response, %{state | correlation_id: state.correlation_id + 1}}
       end
 
+    state = update_metadata(state)
+
```
jbruggem: here, we update the metadata when we delete topics

lib/kafka_ex/server_0_p_10_and_later.ex
```diff
@@ -273,6 +275,8 @@ defmodule KafkaEx.Server0P10AndLater do
...
```
jbruggem: here, we update the metadata when we create topics

test/integration/server0_p_10_and_later_test.exs
joshuawscott: This actually makes a lot more sense - this could be a use case in production code, and it would be annoying to make users insert a wait for this to be updated.

jbruggem:
> @jbruggem One question, otherwise just need to rebase and we're good to go.

Done, thanks !

joshuawscott:
@jbruggem we'd love to have you over on the elixir slack channel #kafkaex - any chance you could join? https://elixir-slackin.herokuapp.com/

jbruggem:
> @jbruggem we'd love to have you over on the elixir slack channel #kafkaex - any chance you could join? https://elixir-slackin.herokuapp.com/

With pleasure ! Thanks for offering.

PR #356, opened 2019-07-20
Kayrock integration - Compatibility with existing API up to Consumer Group usage

dantswain:
This still needs some clean up for sure, but it actually passes all of the tests now so I wanted to get it in front of :eyes: to get feedback on the approach.

The basic idea here is to create a new server implementation that will leverage Kayrock (https://github.com/dantswain/kayrock which is in turn based on https://github.com/klarna/kafka_protocol) for the Kafka protocol instead of KafkaEx's piecemeal protocol implementation which has been stuck at V0 forever.

I think the right ultimate thing to do is to create a new API.  The existing server.ex implementation does a lot of request/response manipulation within the server, which makes it very hard to extend the protocol.  The new server (`server_kayrock.ex`) should ultimately only know about how the metadata messages are constructed so that it can maintain cluster metadata.  Furthermore, the existing API just has a lot of warts that could be cleaned up.  This is a long-term idea; it would be a 1.0 release with proper rollout.

My approach here is to sandbox the "new" API (which is incomplete) into a `KafkaEx.New` namespace and start with a server implementation that is _compatible_ with the existing `KafkaEx.*` API.  That much is present in this PR.  I have ported almost every integration test to `test/integration/kayrock/compatibility_*`.

Things that are NOT included in this PR:
* Documentation and specs - I'd like for the API to stabilize a little first
* Using an actual release of Kayrock - It points at my github because I've had to make some updates while working on KafkaEx.  I will try to do this before merging.
* Consumer group implementation - See below.

Breaking changes in this PR:
* Drop support for Kafka < 0.11 - Minimum for V4 of the metadata protocol
* Drop support for Elixir 1.1 and 1.2 - These are ancient now and don't have the `with` construct.

For consumer group implementation, I think we will need to make a small change to the actual consumer group code and I wanted to make that in a separate PR.  Basically, when `server_kayrock` starts up, it doesn't request metadata for every topic, which breaks consumer group startup since we need to know metadata about.  This is a slightly selfish change since I noticed that on clusters with a large number of topics the initial metadata sync can time out.


dantswain:
@joshuawscott @bjhaid I think this is finally ready for review 😓   I know it's a lot but I'm not sure how to break it down into smaller parts.

The plan I have in mind is to
* get this merged
* finish compatibility support for consumer groups (will require small changes to existing consumer group implementation)
* add support for v2 record batches in the legacy api via ServerKayrock
* add support for offset commits via zookeeper via ServerKayrock
(at this point I think KafkaEx supports the most-demanded features of the newer message versions)
* start fleshing out the `KafkaEx.New.KafkaExAPI`
* get people to test the new API
* eventually invert so that `KafkaEx.New.KafkaExAPI` is just `KafkaEx` and the existing API becomes something like `KafkaEx.Legacy.KafkaExAPI` and is marked for future deprecation - this would be e.g., a 1.0.0 release
* eventually after that drop support for the legacy api completely

bjhaid:
@dantswain ACK, I am traveling this week and cannot review till maybe next week

dantswain:
@bjhaid no worries.  If you have limited time, try to focus on the general approach - not much more than what I have put in my notes.

lib/kafka_ex/server_kayrock/state.ex
```diff
@@ -0,0 +1,148 @@
...
+    cluster_metadata: %ClusterMetadata{},
+    correlation_id: 0,
+    consumer_group_for_auto_commit: nil,
+    metadata_update_interval: nil,
+    consumer_group_update_interval: nil,
+    worker_name: KafkaEx.Server,
+    ssl_options: [],
+    use_ssl: false,
+    api_versions: %{},
+    allow_auto_topic_creation: true
```
joshuawscott: This seems like it should be somewhere else, since it's a broker setting. We don't hold other broker settings that are even more important, like the maximum message size.

lib/kafka_ex/new/kafka_ex_api.ex
```diff
@@ -0,0 +1,88 @@
...
+    case error_code do
+      0 -> {:ok, offset}
+      _ -> {:error, Kayrock.ErrorCode.code_to_atom(error_code)}
+    end
+  end
+
+  @doc """
+  Get topic metadata for the given topics
+
+  Fetches metadata from the server if necessary.  Set allow_topic_creation to
```
joshuawscott: Do we allow auto-creating topics now if we request the metadata from them?

It also seems like this always fetches metadata from the server?

lib/kafka_ex/new/kafka_ex_api.ex
```diff
@@ -0,0 +1,88 @@
...
+  """
+  @spec correlation_id(client) :: {:ok, correlation_id}
+  def correlation_id(client) do
+    GenServer.call(client, :correlation_id)
+  end
+
+  @doc """
+  Set the consumer group name that will be used by the given client for autocommit
+  """
+  @spec set_consumer_group_for_auto_commit(client, consumer_group_name) ::
```
joshuawscott: I'm not sure about having a client know about the consumer group, or having the top-level API have hooks into that. It seems like we should try to keep all the consumer group code isolated, given the complexities of it.

lib/kafka_ex/server_kayrock/state.ex
dantswain: This is a field in the metadata request as of v4.  It's `AND`ed with the cluster setting IIUC.  I think long-term I'd be completely OK with disabling this as a setting, but a lot of our existing tests depend on auto topic creation (which is not great).  So it's either allow it now and plan to remove it once we drop compatibility, or rework a ton of the compatibility tests.  I think I'm kind of OK with either approach but leaving it for now is a lot less work.

lib/kafka_ex/new/kafka_ex_api.ex
dantswain: Yes, we auto-create by default in the existing implementation.  Before metadata request v4 that was the default.

This does always fetch from the brokers.  It seems like the only use cases for this call are either tests or ops-type things, so it doesn't make sense to me to just return the cached metadata.  Especially since we may or may not have the requested topic(s) cached.  (If you didn't catch it, I disabled fetching metadata for _every_ topic - I was actually finding that to cause timeouts with our cluster at work because it has so many topics.  The kayrock client implementation fetches metadata on-demand for topics.)

lib/kafka_ex/new/kafka_ex_api.ex
dantswain: This is another thing where I think I'm ok with removing it eventually but we kind of need it right now for compatibility.  This is the equivalent of the (poorly named) `consumer_group` field in the existing implementation states, which is really only used for auto-commit.  I didn't want to tie the kayrock client to a single consumer group like that, so I renamed the field to make it more obvious what it's used for.  I fully intend to use the existing consumer group implementation and keep it out of the client as much as possible.

lib/kafka_ex/new/kafka_ex_api.ex
```diff
@@ -0,0 +1,88 @@
+defmodule KafkaEx.New.KafkaExAPI do
+  @moduledoc """
+  This module interfaces Kafka through the ServerKayrock implementation
+
+  This is intended to become the future KafkaEx API
+  """
+
+  alias KafkaEx.ServerKayrock
```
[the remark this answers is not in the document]
dantswain: Actually what about just `KafkaEx.Client`?  The goal here is to ultimately have one client that will work with the protocol as it changes.

dantswain:
Not sure why sourcelevel just now found a bunch of issues with the markdown files.

lib/kafka_ex/new/kafka_ex_api.ex
joshuawscott: ok, makes sense. I guess just a tweak to the comment so that it's more clear that it's always doing a fetch from the brokers for the metadata for the topic(s) and I'm 👍

lib/kafka_ex/new/kafka_ex_api.ex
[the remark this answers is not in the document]
joshuawscott: 👍

lib/kafka_ex/new/cluster_metadata.ex
```diff
@@ -0,0 +1,268 @@
...
+    * `{:consumer_group, consumer_group_name}` - the controller for a given
+      consumer group
+  """
+  @type node_selector ::
+          KafkaExAPI.node_id()
+          | :controller
+          | :random
+          | {:topic_partition, KafkaExAPI.topic_name(),
+             KafkaExAPI.partition_id()}
+          | {:consumer_group, KafkaExAPI.consumer_group_name()}
```
jbruggem: I like this.

One caveat: when reading the code using functions that take a `node_selector`, it wasn't easy to understand for me that it was indeed a node selector (for example: `{:topic partition, topic, partition}`). 

Suggestion: have a module that helps build the selector. For example:

```elixir
defmodule Node do
   def for_topic_partition(topic, partition), do: {:topic_partition, topic, partition}
   def random(), do: :random
   # etc...
end

# Now we could write :
ServerKayrock.kayrock_call(server, request, Node.for_topic_partition(topic, partition))
```

Maybe it's out of the scope of what you wish to do here, though (also maybe it's overkill ^^).

lib/kafka_ex/server_kayrock.ex
```diff
@@ -0,0 +1,946 @@
...
+
+  def handle_call({:metadata, topic}, _from, state) do
+    updated_state = update_metadata(state, [topic])
+
+    {:reply, Adapter.metadata_response(updated_state.cluster_metadata),
+     updated_state}
+  end
+
+  ######################################################################
+  # compatibility handle_call messages
```
jbruggem: I think it would be interesting to have a seperate GenServer for the compatbilitiy with the existing API, and have that server call a separate GenServer implementing the new API (by sending `:kayrock_request` calls). It would help make sure that there's a clear seperation between code that's there to maintain the old API and the code implementing the new API.

lib/kafka_ex/server_kayrock.ex
```diff
@@ -0,0 +1,946 @@
...
+        0 -> false
+        _ -> true
+      end
+
+    network_timeout = config_sync_timeout(network_timeout)
+
+    {sender, updated_state} =
+      get_sender(node_selector, state, network_timeout, synchronous)
+
+    case sender do
```
jbruggem: I would have expected `sender` to be a verb, since it's a function -- it was a bit confusing

lib/kafka_ex/server_kayrock.ex
```diff
@@ -0,0 +1,946 @@
...
+  @doc """
+  Make a
+  """
+  @spec kayrock_call(
+          KafkaEx.New.KafkaExAPI.client(),
+          map,
+          KafkaEx.New.ClusterMetadata.node_selector(),
+          pos_integer | nil
+        ) :: {:ok, term} | {:error, term}
+  def kayrock_call(
```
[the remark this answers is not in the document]
dantswain: Good point.  I got the `call` semantics from `GenServer.call` but that's not a great idea because it ties the API to the implementation.

lib/kafka_ex/new/cluster_metadata.ex
dantswain: Not overkill, I like it.

lib/kafka_ex/server_kayrock.ex
dantswain: Hm, I think I follow.  You would have a GenServer implementation that's compatible with the old API that does all of the adapting work but then just relays to the new client for the work?  That would definitely keep things separated but I'm worried that adding that extra server layer will complicate things.  With the implementation here I was going for a drop-in replacement, especially with the ConsumerGroup code which is much newer and cleaner than the old KafkaEx API and I hope to keep most of that untouched.

lib/kafka_ex/server_kayrock.ex
dantswain: hm good point, I'll try to think of something better.

dantswain:
> Question: would it be worth it to expose a "message-level" API for those who wish to use those directly ? In other words, expose all the messages of the Kafka protocol as functions one can call ? (My understanding is that it would be a thin wrapper around Kayrock). Of course, that would mean using those same functions to implement a higher-level KafkaEx API

This is exactly what I'm aiming for :)  The new API should mostly just be a convenience layer on top of the Kafka API message layer, and I'd definitely expose that directly for "power" users / future-proofing.  I'm a big believer in the the idea that the best abstractions have a trap door to let you go around them if you need to.

dantswain:
Actually rereading your message @jbruggem maybe we mean something slightly different but similar in spirit?

It sounds like you're suggesting something like `KafkaExAPI.produce(%Kayrock.Produce.VX.Request{...})`, but I mean to provide a simplified `KafkaExAPI.produce(messages=..., topic=..., partition=...)` that just creates a `%Kayrock.Produce.VX.Request{...}` and deals with the response message as well.  But this would be done by calling through a `KafkaExAPI.client_request(Node.partition_topic(partition, topic) ,%Kayrock.Produce.VX.Request{...})` - i.e., the client is only responsible for knowing how to send/receive the request and doesn't care what the message is.  I DO intend to make `KafkaExAPI.client_request(node_selector, request)` a public and documented API function, which would allow anyone to send any Kafka API request directly and sidestep the convenience layer.

I think at the end of the day we're both saying it would be good to allow people to send those raw request messages?

jbruggem:
@dantswain yes, I think we're saying mostly the same things. I agree with the fact that the low-level API should be simple (so `KafkaExAPI.produce(messages=..., topic=..., partition=...)`), because I believe KafkaEx shouldn't expose Kayrock to users, and I believe a good API should not require users to build specific Structs to be able to use it (so I think args should be plain data types).

Where we might differ in opinion is that I think this low-level API could be done without any process, as a simple module (but I might be wrong, because my knowledge of the Kafka protocol is incomplete). It would work in a similar manner than Tesla does, by giving the user of the API the responsibility of providing the "state object/client instance" to Tesla whenever they call a function of that API.

Would that be feasible ? Do you think it makes sense ?

For example:

```elixir
client = KafkaExAPI.connect(broker_urls)

{:ok, _} = client |> KafkaExAPI.produce(messages=..., topic=..., partition=...)

```

> I'm a big believer in the the idea that the best abstractions have a trap door to let you go around them if you need to.

Agreed, and it's important for things to be clear for the user (i.e: am I using high-level or low-level)

>  I DO intend to make KafkaExAPI.client_request(node_selector, request) a public and documented API function, 

I think it's a good idea. With a clear message of "Here be dragons" in the doc :).

> I think at the end of the day we're both saying it would be good to allow people to send those raw request messages?

yes !

lib/kafka_ex/server_kayrock.ex
```diff
@@ -0,0 +1,946 @@
...
+      | topics: topics,
+        allow_auto_topic_creation: state.allow_auto_topic_creation
+    }
+
+    {{ok_or_err, response}, state_out} =
+      kayrock_network_request(metadata_request, :any, state)
+
+    case ok_or_err do
+      :ok ->
+        # THIS WILL PROBABLY NOT WORK
```
jbruggem: Should we be worried ? ^_^

lib/kafka_ex/server_kayrock.ex
jbruggem: I'm not sure I understand everything in your answer :/. 

>  especially with the ConsumerGroup code which is much newer and cleaner than the old KafkaEx API and I hope to keep most of that untouched.

You mean you want to be able to keep the new code you wrote for when you'll write the new API ?

Anyway, here's what I'm worried about, It's probably better to start with that: I'm worried that `ServerKayrock` will contain both:
1. code that is needed in the new, future `KafkaEx.Server`
2. code that is there to implement the legacy `KafkaEx.Server` API.

That worries me because:
- it could taint the logic of the code that needs to be kept, because someone will tweak it to "make it work the the legacy API"
- it could make it hard to discern what must be removed once the legacy API is removed.

My suggestion comes from that worry. To prevent leakage from the legacy API to the new API, I figured having a proxy process would help -- all the "good new code" we want to keep would be kept separately from the code that's only needed to maintain compatibility.

Alternative suggestion: 
1. extract all the logic out of the GenServer. 
2. make two GenServers, one for compatibility and one for the new API
3. use the common, extracted logic in both

Maybe my worries are not justified -- maybe I just don't understand well enough how it works ?

dantswain: Ah I see.  Most of the logic in `ServerKayrock` is intended to be the new logic that will stick around.  It's based heavily on the existing `KafkaEx.Server` logic because that's sort of battle-tested at this point, but with some refactoring to make it more maintainable and also trying to keep as much of the request processing logic outside of the server (this tends to be what makes keeping up with the Kafka API difficult).

I'll try to think of some way to make the compatibility code more self-contained so that it's easier to extract later.

lib/kafka_ex/server_kayrock.ex
dantswain: Haha some of this is comments left over from me transposing the old code.  My guess is I just forgot to delete the comments - I'll do some double checking.

dantswain:
Hm.  The GenServer in the client serves to maintain our picture of the state of the cluster.  Most requests need to be routed to the appropriate broker.  We hold that information in the GenServer state and the routing logic is part of handler callbacks.  It makes sense to me for that to be in a process.  We also do things like schedule timers to periodically update the metadata.  It also makes it so that individual broker connections are isolated from the calling code and should get restarted automatically.

I think it's _possible_ to do this without that process, but I think it would be more cumbersome.  When you do, e.g., `{:ok, client} = KafkaEx.connect(...)`, `client` would have to be a struct containing the cluster metadata, and that object would potentially have to get updated on every request.  This wouldn't let the user share a broker connection very easily amongst multiple parts of their code (vs, say, registering the process, or even just passing a single `pid` around).


Sorry it's taken me a few days to get back to you - I've had a sick kid and we're getting ready to move :sweat:

lib/kafka_ex/server_kayrock.ex
```diff
@@ -0,0 +1,946 @@
...
+  def start_link(args, :no_name) do
+    GenServer.start_link(__MODULE__, [args, nil])
+  end
+
+  def start_link(args, name) do
+    GenServer.start_link(__MODULE__, [args, name], name: name)
+  end
+
+  @doc """
+  Make a
```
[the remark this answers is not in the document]
dantswain: Finished my sentence :joy:

lib/kafka_ex/server_kayrock.ex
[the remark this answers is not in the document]
dantswain: Renamed to `send_request`

lib/kafka_ex/new/cluster_metadata.ex
dantswain: I added a `NodeSelector` module.  I ended up using a struct with a `:strategy` field in order to make it easier to make clear pattern matches.

lib/kafka_ex/server_kayrock.ex
dantswain: Renamed to `get_send_request_function`/`send_request`

lib/kafka_ex/server_kayrock.ex
dantswain: Ya I checked with the coverage tool, it works.  Just an old comment.

lib/kafka_ex/server_kayrock.ex
dantswain: I moved all of the compatibility `handle_call` clauses into a mixin module - does that work?  I don't want to expend _too_ much energy or make the code a lot more complex to avoid something that we should catch in code review during the transition period.  I'm open to discussion on that, though.

lib/kafka_ex/new/kafka_ex_api.ex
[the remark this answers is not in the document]
dantswain: Renamed to `KafkaEx.New.Client`

dantswain:
@joshuawscott @jbruggem I think I've addressed all of the individual feedback requests.  I marked conversations as "resolved" mostly just as a way for me to keep track of what requests were still outstanding - feel free to un-resolve them.  Renaming the `ServerKayrock` module may end up making the comment history sort of messy 🤷‍♂ 

There's one flakey test (getting topic metadata) that I'd ideally like to iron out but that seems to be an ongoing battle.

jbruggem:
> Sorry it's taken me a few days to get back to you - I've had a sick kid and we're getting ready to move

No worries ! I hope your kid is better now. All the best for the installation in your new home.

> I think I've addressed all of the individual feedback requests. 

I'll take a look as soon as possible.

jbruggem:
> I'll take a look as soon as possible. 

Done. I like what you did, it makes things clearer ! 

:+1:

PR #359, opened 2019-08-22
Handle consumer groups with new client

dantswain:
This ended up being surprisingly easy.  Where I got hung up before is
that the new client doesn't know about any topics by default (fetching
all topics for a large cluster can cause timeouts), but it needs to know
about the topics that belong to the consumer group.  So I added an
`initial_topics` setting for the new client's initialization and pass
that in from the consumer group set up.  The old client will just ignore
this setting.


lib/kafka_ex/gen_consumer.ex
```diff
@@ -698,7 +703,7 @@ defmodule KafkaEx.GenConsumer do
       end
 
     case response do
-      %{last_offset: nil, message_set: []} ->
```
dantswain: I'm not sure why we were matching on this before.  The new client passes the actual last offset even if the message set is empty.

lib/kafka_ex/new/client.ex
```diff
@@ -22,7 +22,9 @@ defmodule KafkaEx.New.Client do
   use GenServer
 
   @type args :: [
-          KafkaEx.worker_setting() | {:allow_auto_topic_creation, boolean}
+          KafkaEx.worker_setting()
+          | {:allow_auto_topic_creation, boolean}
+          | {:iniital_topics, [binary]}
```
joshuawscott: iniital -> initial

dantswain: Fixed.  I had to actually move this to `KafkaEx.worker_settings` to get dialyzer to pass.

Issue #358, opened 2019-08-15
Add Topic and Partition ID to Protocol.Fetch.Message


yogodoshi:
On the [Message struct](https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/protocol/fetch.ex#L46) there is the offset field but not the topic or the partition ID. It would be very useful for me if them were there as well, here is why:

```
topic_name
|> retrieve_partitions_ids
|> create_streams(topic_name)
|> Flow.from_enumerables()
|> Flow.map(&prepare_for_partition/1)
|> Flow.partition(key: {:elem, 0}, window: periodic_window)
|> Flow.reduce(&accumulator_tuple/0, &save_kafka_message_to_ets/2)
(more Flow stuff)
|> Flow.on_trigger(&send_to_redshift/3)
|> Flow.run()
```

I have a Flow with a `KafkaEx.stream` for each partition and down the road, I don't really have a way to know from which partition each message is from, which prevents me from committing the offset.

I would think that if this information would be great to have in my scenario, it would probably also be useful to others as a Map-Reduce is probably frequently used with `KafkaEx.stream` 🤔 

If you think this change makes sense let me know and I can try opening a PR implementing it :)

joshuawscott:
Fixed in #360

PR #360, opened 2019-08-27
Include topic and partition in message struct

thiamsantos:
## Motivation

- In our current use case, we want to consume multiple topics, with the consumer group supervisor. To do that we need a way to identify the message to each topic reaching the handler.

## Proposed solution

- Include the topic key in the `Protocol.Fetch.Message ` struct of the response.
- Include the partition key in the `Protocol.Fetch.Message ` struct of the response.

## Related

#358

### DIFF
--- a/lib/kafka_ex/protocol/common.ex
+++ b/lib/kafka_ex/protocol/common.ex
@@ -22,7 +22,7 @@ defmodule KafkaEx.Protocol.Common do
         mod
       ) do
     struct_module = Module.concat(mod, Response)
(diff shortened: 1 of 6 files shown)


thiamsantos:
Thanks for the review :heart: 

@dantswain @joshuawscott Do you plan to release a new hex version with these changes soon?

joshuawscott:
@thiamsantos my apologies for the long release delay - I'm going to have a new release today or tomorrow (depending on how long it takes to get through the release notes)

PR #364, opened 2019-09-11
Allow specifying API version for produce/fetch with new client, support timestamps

dantswain:
The produce and fetch commands now accept an `api_version` option that will be used _only_ with the new Kayrock-based client to send the appropriate-versioned message.

This is important for compatibility with newer versions of Kafka that use the more efficient RecordBatch message storage format.  The existing KafkaEx code always sends a v0 fetch request, which will force the broker to convert fetched messages, which can lead to very high CPU load on the broker.

I also added support for timestamps with the appropriate api versions.

I added a pile of test cases for various combinations of produce/fetch versions, compression settings, and timestamp settings.


lib/kafka_ex/protocol/produce.ex
```diff
@@ -44,9 +47,10 @@ defmodule KafkaEx.Protocol.Produce do
     - key: is used for partition assignment, can be nil, when none is provided
     it is defaulted to nil
     - value: is the message to be written to kafka logs.
+    - timestamp: timestamp (`kafka_version: "kayrock"` ONLY)
     """
-    defstruct key: nil, value: nil
-    @type t :: %Message{key: binary, value: binary}
+    defstruct key: nil, value: nil, timestamp: nil
+    @type t :: %Message{key: binary, value: binary, timestamp: integer}
```
esambo, line 53: shouldn't the [timestamp](https://github.com/kafkaex/kafka_ex/issues/261#issuecomment-530973125) be optional?
```suggestion
    @type t :: %Message{key: binary, value: binary, timestamp: integer | nil}
```

PR #366, opened 2019-09-19
Allow specifying versions for offset commit and fetch

dantswain:
I also made a small change in client.ex to allow the node_id node selector to work (and any other that can pass through straight to ClusterMetadata).

Also FWIW I found that if you commit to zookeeper (v0) and try to fetch the same consumer group from kafka (v > 0), you get `:topic_or_partition_not_found`.  If you commit to kafka (v > 0) and try to fetch from zookeeper, you get an offset of -1.  So people who have already been committing to zookeeper will need a migration plan.

### DIFF
--- a/lib/kafka_ex/new/adapter.ex
+++ b/lib/kafka_ex/new/adapter.ex
@@ -343,18 +343,26 @@ defmodule KafkaEx.New.Adapter do
     }
   end
 
-  def offset_fetch_request(request, client_consumer_group) do
-    consumer_group = request.consumer_group || client_consumer_group
+  def offset_fetch_request(offset_fetch_request, client_consumer_group) do
+    consumer_group =
+      offset_fetch_request.consumer_group || client_consumer_group
 
-    {%Kayrock.OffsetFetch.V0.Request{
-       group_id: consumer_group,
-       topics: [
-         %{topic: request.topic, partitions: [%{partition: request.partition}]}
-       ]
+    request =
+      Kayrock.OffsetFetch.get_request_struct(offset_fetch_request.api_version)
+
+    {%{
+       request
+       | group_id: consumer_group,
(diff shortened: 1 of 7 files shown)


lib/kafka_ex/new/adapter.ex
```diff
@@ -384,32 +392,77 @@ defmodule KafkaEx.New.Adapter do
...
+                partition: offset_commit_request.partition,
+                offset: offset_commit_request.offset,
+                metadata: ""
+              }
+            ]
+          }
+        ]
+    }
+
+    request =
```
joshuawscott: This request is from Kayrock.OffsetCommit.get_request_struct/1, which we pass the api version into. It seems like it should handle this case statement as well, otherwise it seems like there's no point in actually passing in the api version?

dantswain: I'm not sure I understand your question.  I'll try to answer, but not 100% sure it'll be an answer to what you're actually asking.

We have to get the request struct from Kayrock so that it's the right struct.  Each version of each message is a different struct - e.g., `%Kayrock.OffsetCommit.V0.Request{}`.  The case statement here (which, yeah, is kind of ugly... but I didn't see any simple way to clean it up significantly without losing readability) populates fields of the struct depending on what version it is (since different versions can have different fields).  Kayrock is sort of intentionally not concerned with knowing how to populate the fields - it just knows what fields should be in what structs and how to serialize/deserialize them.

The weird/stupid thing is that the OffsetCommit request had a timestamp field in v1 that got removed in v2.  Otherwise this would probably be as simple as checking if version >= 1 and populating the generation_id and member_id fields which are common in v >= 1.

joshuawscott: Gotcha, makes sense.

Issue #226, opened 2017-08-30
Consumer groups do not rebalance automatically when partitioning changes


dantswain:
The following can occur:

* A topic exists with, say, 4 partitions.
* A KafkaEx consumer group begins consuming from those partitions.
* The topic is reconfigured to have, say, 8 partitions.
* The KafkaEx consumer group will continue to only consume from the initial 4 partitions until a rebalance is forced (e.g., a worker restarts/joins/leaves).

For now this can be managed easily enough by manually restarting a worker, but it should be done automatically.  The leader can monitor the metadata for changes relatively easily.  I think the main "challenge" will be detecting which changes require a rebalance.  The design doc (https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Client-side+Assignment+Proposal) doesn't seem to address this condition specifically, but it could be construed as being covered by the duties of the leader:

> The leader of the group can enforce its own policy for controlling the rate of rebalancing. It doesn't have to rebalance after every metadata change, but can "batch" changes together to reduce the impact of metadata churn.

PR #371, opened 2019-10-06
Expose rebalance function

thiamsantos:
### Motivation

Consumer groups do not rebalance automatically when partitioning changes, so we need a way to trigger this rebalance.

### Proposed solution

We already have a internal rebalance function, by exposing it to consumer group API we may be able to trigger a rebalance whenever we think is appropriate.

Right now, in my setup I have been using a genserver that periodically checks for the number of partitions of the given topic, if the number changes the rebalance is triggered so the consumers.

I may need some help with the unit tests for this function. Also, I don't know if this functions is going in the right direction, if that's so we may want to also include in the library itself a poller like the one I have been using, and trigger the rebalance automatically inside the library.

Related to #226

### DIFF
--- a/lib/kafka_ex/consumer_group.ex
+++ b/lib/kafka_ex/consumer_group.ex
@@ -231,7 +231,11 @@ defmodule KafkaEx.ConsumerGroup do
     call_manager(supervisor_pid, :am_leader, timeout)
   end
 
-  @spec rebalance(Supervisor.supervisor(), timeout) :: boolean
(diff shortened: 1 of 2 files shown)


dantswain:
Hi @thiamsantos thanks for looking into this!  I'm a little surprised that Kafka doesn't automatically invalidate the consumer group when this happens and trigger a rebalance automatically.  Is it possible that is happening and we're just not catching it correctly?

PR #378, opened 2019-10-13
Add `dwell_on_empty` option for GenConsumer

dantswain:
I noticed that I was getting timeouts when testing consumer groups
locally.  I think this is due to the test consumer overwhelming my test
kafka cluster (running in docker).  I noticed that when we handle an
empty message set, we immediately poll again for new messages.  This is
probably what many, if not most, users want but I can also see where it
could be an issue for some users.  I added an option for GenConsumer
(which is passed through by ConsumerGroup options) to specify a number
of milliseconds to dwell if we receive an empty message set.  The
default is zero, which is equivalent to the existing behavior.

### DIFF
--- a/lib/kafka_ex/gen_consumer.ex
+++ b/lib/kafka_ex/gen_consumer.ex
@@ -397,6 +397,7 @@ defmodule KafkaEx.GenConsumer do
       :last_commit,
       :auto_offset_reset,
       :fetch_options,
+      :dwell_on_empty,
       :api_versions
     ]
   end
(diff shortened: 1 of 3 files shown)


joshuawscott:
I think this can be done via the max_wait_time in the Fetch protocol. We expose this as `wait_time` in `%Fetch.Request{}`. 
https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-FetchRequest

PR #388, opened 2019-11-27
Configurable client id

davidrusu:
Kafka lets us set produce and consume quotas to prevent badly behaved consumers/producers from eating up all the bandwidth of a broker, read more about it here: https://kafka.apache.org/documentation/#design_quotas

Kafka support's enforcing quota's by `client_id`, `user_id`, or by `user_id,client_id` pairs, but we don't use ACL's and as such we mainly care for Client ID based quota calculations.

KafkaEx currently does not support this use case since it hard codes the client id to `kafka_ex`. All services that interact with kafka will be accounted for under the same Quota.

This PR exposes the client id to be configurable by library users:
``` elixir
config :kafka_ex,
    client_id: "badly_behaved_service"
```

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -33,6 +33,8 @@ config :kafka_ex,
   #    consumer groups, set this to :no_consumer_group (this is the
   #    only exception to the requirement that this value be a binary)
   consumer_group: "kafka_ex",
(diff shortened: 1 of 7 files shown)


dantswain:
I think ideally this would be something that could be set per-worker, but that can be deferred to another PR.

PR #391, opened 2019-12-25
support snappyer

getong:


.travis.yml
```diff
@@ -18,7 +16,6 @@ matrix:
       otp_release: 20.2
     - elixir: 1.4.5
       otp_release: 19.3
-      env: COVERALLS=true CREDO=true
```
joshuawscott: Removing this line disables credo. It should be put into one of the matrix builds, probably the latest version.

README.md
```diff
@@ -474,3 +474,10 @@ an invite via [http://bit.ly/slackelixir](http://bit.ly/slackelixir).
 The Slack channel is appropriate for quick questions or general design
 discussions.  The Slack discussion is archived at
 [http://slack.elixirhq.com/kafkaex](http://slack.elixirhq.com/kafkaex).
+
+## default snappy algorithm use snappyer package
```
joshuawscott: it looks like the default is actually snappy - should we change this to say "Changing the default snappy library to snappyer"?

getong:
The credo build faild, so I delete it.

bjhaid:
@getong 

> The credo build faild, so I delete it.

please don't delete it, instead fix the failures/warnings

getong:
@bjhaid @joshuawscott 
The build test is passed,  please take a review of it.

lib/kafka_ex/server_0_p_10_and_later.ex
```diff
@@ -296,54 +297,54 @@ defmodule KafkaEx.Server0P10AndLater do
...
-
-        update_consumer_metadata(
-          %{state | correlation_id: state.correlation_id + 1},
-          retry - 1,
-          response.error_code
-        )
-    end
-  end
+##  defp update_consumer_metadata(
+##         %State{consumer_group: consumer_group} = state,
```
bjhaid: Why is this getting commented out?

getong:
``` elixir
defdelegate update_consumer_metadata(state, num, error_code), to: Server0P8P2
```
Just because the `credo` complains the three modules use the same function, the three functions definition are the same.

lib/kafka_ex/server_0_p_8_p_2.ex
```diff
@@ -213,7 +213,7 @@ defmodule KafkaEx.Server0P8P2 do
   defp update_consumer_metadata(state),
```
joshuawscott: This head is still private, is that intended? it's strange for a private function to wrap a public one

lib/kafka_ex/server_0_p_10_and_later.ex
joshuawscott: This seems fine, but the code needs to be deleted rather than commented out

lib/kafka_ex/server_0_p_8_p_2.ex
getong: This private function is used by their own module, do it need to be changed?

lib/kafka_ex/server_0_p_10_and_later.ex
getong: OK, it is deleted.

README.md
bjhaid: @getong please consider @joshuawscott recommendation here thanks!

.travis.yml
```diff
@@ -1,13 +1,13 @@
 language: elixir
-branches:
```
dantswain: Can you please revert this?

config/config.exs
```diff
@@ -65,6 +65,8 @@ config :kafka_ex,
     certfile: File.cwd!() <> "/ssl/cert.pem",
     keyfile: File.cwd!() <> "/ssl/key.pem"
   ],
+
+  snappy_module: :snappyer,
```
dantswain: This makes it so that the default in dev will be snappyer, and a lot of people will just copy this into their own projects which will make the documentation confusing for them.  Please either change it to `:snappy` it or update the documentation.

config/test.exs
```diff
@@ -1,3 +1,5 @@
 use Mix.Config
 
+config :kafka_ex, snappy_module: :snappyer
```
dantswain: Again, please don't override this unless we plan to make snappyer the default.  You can (and we should) test snappyer by wrapping a test where we change the config.

.travis.yml
```diff
@@ -1,13 +1,13 @@
...
-  only:
-    - "master"
+
 cache:
   directories:
     - deps
     - _build
 matrix:
   include:
+    - elixir: 1.9.4
```
dantswain: This is a good thing to do but it doesn't seem directly related to the snappyer change.  If you want to make this change, please make a separate pull request.

lib/kafka_ex/server_0_p_10_and_later.ex
```diff
@@ -40,6 +39,7 @@ defmodule KafkaEx.Server0P10AndLater do
 
   defdelegate kafka_server_consumer_group_metadata(state), to: Server0P8P2
   defdelegate kafka_server_update_consumer_metadata(state), to: Server0P8P2
+  defdelegate update_consumer_metadata(state, num, error_code), to: Server0P8P2
```
dantswain: This doesn't seem to be necessary for the snappyer change.  Please make it a separate pull request.

mix.exs
```diff
@@ -40,11 +40,12 @@ defmodule KafkaEx.Mixfile do
   defp deps do
     main_deps = [
       {:kayrock, "~> 0.1.9"},
-      {:credo, "~> 0.8.10", only: :dev},
+      {:credo, "~> 1.1", only: :dev},
```
dantswain: Please make this a separate pull request.

config/config.exs
getong: The doc is changed.

config/test.exs
getong: This is forced to use in the testenvironment and the ci environment.

getong:
@dantswain 
The main reason is the `snappy` package does not maintain well, even the github website does not exist, when it compile with erlang 22, it has a lot of warning, and the `snappyer` package does maintain well. Even by the download number, the  `snappyer` does have 92344, but `snappy` only has 3257, which is quite different, you can find this in https://hex.pm/packages?search=snappy&sort=recent_downloads . Both packges have the same methods. By adding this option, we can change the default package now and later.

dantswain:
@getong I'm sorry it's taken me so long to get back to you on this, I've been very busy.

I think in the long term we should simply switch to using snappyer.  I'd be a little concerned with doing that right away because it could be a surprise breaking change (though easily fixed).  OTOH I don't like the added complexity of supporting two implementations.  

@joshuawscott @bjhaid @jbruggem I think we have a few options:

1. Leave everything as-is for now and switch to snappyer for a 1.0 release.  This means no code complexity to support both implementations, but delays the cut over.
2. Switch to snappyer now.  This is the cleanest, rip-off-the-bandaid-est approach, but it introduces the possibility for surprise breaking changes.  We can reduce the impact with documentation, and the solution to the breaking change is very simple (change a hex dependency), but it is still potentially a breaking change.  Strictly speaking, semver doesn't obligate us not to introduce breaking changes sub-1.0 🤷‍♂ 
3. Take @getong's changes here which switches the default implementation to snappyer.  This will probably cause all of the same complications as (2) since it would require a configuration change to revert to using snappy_erlang_nif.  I think (2) is preferable to this because both are breaking changes and the solutions for end-users are similarly simple but (2) is a better path forward.
4. Modify @getong's changes so that snappyer is not the default and switch to snappyer at 1.0.  This is the most user-friendly path forward but at the cost of code complexity now.

@getong If we go with either option (3) or (4), we will need to add tests so that both snappy and snappyer are tested.  I can help you accomplish this.

getong:
Make your choice and I will do it.

jbruggem:
My opinion: go for option 4., since it seems supporting both doesn't involve a lot of code. It means:
- we don't break current behaviour
- we get to see how the new lib behaves before making it the default (via the test + via @getong and other who will probably activate the new lib in their deployments)

getong:
@jbruggem 
As you suggested, the code has been changed.

config/test.exs
jbruggem: This changes the config for *all* the tests, meaning we're not protected against regressions when using the old snappy lib.

This is problematic, because it means that the test suite wouldn't cover *at all* the default config we're shipping (i.e. the old snappy lib).

I suggest modifying the tests that cover compression so that they run twice, once with the default config and one with the new snappy module.

You can change the Application config for a given test (and restore it afterwards) by using `Application.put_env`.

bjhaid: do we even need to test snappyer? is this a case where we need to define a compression behavior(interface in other languages)  such that this is swappable in the future without any modification to kafka_ex?

getong: Not test snappyer

jbruggem: > do we even need to test snappyer?

If you're talking about the lib itself, we don't. We do need to test the it integrates correctly, though right ? What I mean is: we do need to test that *our* code, when calling snappyer, works as expected.

Also, to re-iterate my earlier point: 
- most of our tests should use as much as possible the default config for the tests (unless we really need to), do be consistent with what we ship in a release.
- whenever there are several configs, we should try to have at least one test checking that the changed config behaves correctly.

>  is this a case where we need to define a compression behavior [...] ?

Interesting idea ! I don't know much about compression in Kafka, so I don't know how useful that would be.

jbruggem: Addendum: the reason why I would recommend having one test with the new snappyer library is that, if we wish to make it the default for KafkaEx 1.0, we must have at least *some* experience with it. Including a test in the CI for snappyer allows us to get a bit of that.

bjhaid: Sorry just seeing this, maybe I wasn't clear, my hope is that we don't have to always have people open PRs like this in the future, we provide them with a interface they need to implement and let them go implement the interface for any new library they want to use for compression, I may be too far away from things and my suggestions might just be flat out wrong, but it seems it's just easier to empower people to change the compression library by implementing an interface rather than always opening PR to kafka_ex every time there is a new shiny/better compression library/support in the elixir ecosystem or kafka.

getong:
I have update the test code, and all the test code are passed, but the coveralls upload error, this is not the test code error, I think the code can be reviewed now.
```
Finished in 73.7 seconds
256 tests, 0 failures
Randomized with seed 544250
** (ExCoveralls.ReportUploadError) Failed to upload the report to 'https://coveralls.io' (reason: status_code = 422, body = {"message":"service_job_id (736636358) must be unique for Travis Jobs not supplying a Coveralls Repo Token","error":true}).
    lib/excoveralls/poster.ex:21: ExCoveralls.Poster.execute/2
    (mix) lib/mix/tasks/test.ex:446: Mix.Tasks.Test.do_run/3
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    lib/mix/tasks.ex:54: Mix.Tasks.Coveralls.do_run/2
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:79: Mix.CLI.run_task/2
    (elixir) lib/code.ex:813: Code.require_file/2
The command "./scripts/ci_tests.sh" exited with 1.
cache.2
store build cache
```

Argonus:
@joshuawscott @getong 
Support has been merged in: https://github.com/kafkaex/kafka_ex/pull/437
This one could be closed.

PR #392, opened 2020-01-10
Fix kayrock stream skips first message

pixelvitamina:
With the new Kayrock client, when `KafkaEx.stream` internally fetches and receives a response with an empty message set, it increased the stream's offset by one. 

Thus, it skipped the first message in this case.

### DIFF
--- a/lib/kafka_ex/stream.ex
+++ b/lib/kafka_ex/stream.ex
@@ -53,6 +53,22 @@ defmodule KafkaEx.Stream do
     ######################################################################
     # Main stream flow control
 
+    # if we get an empty response, we block until messages are ready
+    defp stream_control(
+           %{
+             error_code: :no_error,
+             last_offset: last_offset,
+             message_set: []
+           },
+           %KafkaEx.Stream{
+             no_wait_at_logend: false
+           },
+           _offset
+         )
+         when is_integer(last_offset) do
+      {[], last_offset}
+    end
+
     # if we get a response, we return the message set and point at the next
     # offset after the last message
     defp stream_control(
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/stream.ex
```diff
@@ -53,6 +53,22 @@ defmodule KafkaEx.Stream do
...
+             last_offset: last_offset,
+             message_set: []
+           },
+           %KafkaEx.Stream{
+             no_wait_at_logend: false
+           },
+           _offset
+         )
+         when is_integer(last_offset) do
+      {[], last_offset}
```
dantswain: I don't feel super strongly about it, but I wonder if it would be clearer to handle this within the following clause?  Something like

```
next_offset = case message_set do
  [] -> last_offset
  _ -> last_offset + 1
end
{message_set, next_offset}
```

My reasoning is that this is controlling what the next offset should be and not a huge difference in what `stream_control` is doing.  I am very open to using your implementation - I'm just throwing this out there as a possible improvement.

pixelvitamina: good point, thanks, will try!

pixelvitamina: @dantswain Thanks a lot for the fast review! 
I think the reason, I used a separate function clause instead of a case inside the function, was that it only should happen in the case of `no_wait_at_logend: false`. 
I quickly tested your suggestion, but `test "stream is non-blocking with no_wait_at_logend"` would fail.

As that case is similarly handled in a separate stream_control function clause, I would argue that it is fine to handle the above case like that, as well.

```elixir
   defp stream_control(
           %{},
           %KafkaEx.Stream{no_wait_at_logend: true},
           offset
         ) do
      {:halt, offset}
    end
```

But I'm also very open to suggestions.

Issue #398, opened 2020-03-04
Default partitioner does not match Java client


mjparrott:
When using KafkaEx to produce to a topic that is also produced to by an application that is using the default Java client, I noticed that the partitions that were being produced to weren't the same in KafkaEx and the Java client.

The reason appears to be how `umurmur2` compares to the Java client.

In KafkaEx, the hash is calculated like so: https://github.com/kafkaex/kafka_ex/blob/403024e2fdd5a6f3ed340b5a5eea4f9e2faf265f/lib/kafka_ex/utils/murmur.ex#L32
In the default Java client, it is calculated here: https://github.com/apache/kafka/blob/717c55be971df862c55f55d245b9997f1d6f998c/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L59

The difference here is in the value we use to bitwise AND the key hash and the number of partitions. In KafkaEx, it is `0xFFFFFFFF`. In the Java client, it is `0x7fffffff`.

Changing the `0xFFFFFFFF` to `0x7FFFFFFF` seems to make things match. I'm wondering if just changing the `umurmur2` method is the correct approach here? And if so, how would versioning work? Since this would affect anyone using KafkaEx's default partitioner. Happy to make a PR if we have an approach.

mjparrott:
I've made a PR that could possibly PR, it is linked to this issue.

jbruggem:
Good catch @mjparrott :+1:

dantswain:
Closed by #399 thanks!

PR #399, opened 2020-03-04
Change default KafkaEx partitioner to correctly match the Java client

mjparrott:
One way of fixing #398

### DIFF
--- a/lib/kafka_ex/default_partitioner.ex
+++ b/lib/kafka_ex/default_partitioner.ex
@@ -23,6 +23,7 @@ defmodule KafkaEx.DefaultPartitioner do
     request
   end
 
+  # credo:disable-for-lines:50 Credo.Check.Design.DuplicatedCode
   def assign_partition(%ProduceRequest{partition: nil} = request, metadata) do
     case Partitioner.get_key(request) do
       {:ok, nil} ->
--- a/lib/kafka_ex/legacy_partitioner.ex
+++ b/lib/kafka_ex/legacy_partitioner.ex
@@ -24,6 +24,7 @@ defmodule KafkaEx.LegacyPartitioner do
     request
   end
 
+  # credo:disable-for-lines:50 Credo.Check.Design.DuplicatedCode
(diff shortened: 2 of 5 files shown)


dantswain:
This makes sense to me.  @barthez did the original murmur implementation - do you have any sense if the original may have just been a typo?

I am a little concerned about breaking implementation for anyone who is already using the existing algorithm.  Maybe we can provide a legacy partitioner?

If nothing else @mjparrott you'll need to update the test cases, which should be relatively straightforward.

kennethito:
I think providing a legacy partitioner and some breaking change documentation is a great idea.

joshuawscott:
I thought these tests had been copied from the Java tests, but it doesn't look like that's actually the case. Perhaps we should remove these test cases, and use the same ones that the Java implementation uses:
https://github.com/apache/kafka/blob/8ab0994919752cd4870e771221ba934a6a539a67/clients/src/test/java/org/apache/kafka/common/utils/UtilsTest.java#L66-L78

jbruggem:
> I am a little concerned about breaking implementation for anyone who is already using the existing algorithm. Maybe we can provide a legacy partitioner?

Indeed, it's touchy. With kayrock we'll have a major release, so having a breaking change is OK, but we need:
- clear documentation on the matter, in the documentation itself and in the release notes
- a way to activate the old behaviour for those in need of it

joshuawscott:
Breaking changes are allowed in semver before 1.0, so I have no issue with breaking this. This is a bug in any case, the intention and documentation is that the partitioning matches the java client, so if it isn't doing that now, it's broken, and changing it to match is fine, IMO

mjparrott:
I added a `LegacyPartitioner` module which can be used in place of the `DefaultPartitioner` which implements the old behaviour. This modules are mostly just a copy-paste of each other. The only difference is the `umurmur2` function which is called.

It looks like the one of the builds is failing due to the code copy-paste - any thoughts?

joshuawscott:
I'm ok if you put in a comment to disable that check in credo
http://rrrene.org/2017/06/01/credo-config-comments/

PR #403, opened 2020-03-18
Increment attempt number instead of decrement when joining a consumer group

mjparrott:
Two things this fixes:

1. The output of `Unable to join consumer group "consumer_group".  Will sleep 1 second and try again (attempt number x)` should be less confusing, as it would start to become a negative number.

2. `attempt_number` is used to raise an error when `@max_join_retries` is reached. This would never be reached when `attempt_number` is decrementing.


jbruggem:
@dantswain thanks for the reminder.

PR #404, opened 2020-03-19
Use IO lists instead of binary concatenation for protocol requests

### LINKED ISSUE #291: Use iolists when building packets
#290 / #231 Implemented using iolists / iodata when building produce protocol messages. This results in a significant speedup in those parts of the code, especially with larger messages/message sets.

We should use a similar strategy for the other protocol message modules.

mjparrott:
Fixes #291 

Seems like there still was some binary concatenation in the protocol modules as far as I could tell, so I followed up on the work done in #290 and made the request for all other protocol call return iolists instead of binaries as well.


Argonus:
@mjparrott Thank you for help! Cherry-picked & added missing stuff.

PR #409, opened 2020-05-27
Do not update metadata cache on metadata call

richardlarocque:
The call to `metadata` can include a list of topics.  If it does, the
metadata we receive will include metadata only for the specified topics.

We don't want to replace our cache of all-topic metadata with a
topic-specific response.  We should leave the cache as-is.

A slightly more optimal way to do this would be to update the metadata
cache only if the metadata request did not specify any topics, but that
seems a bit more effort than it's worth.

### DIFF
--- a/lib/kafka_ex/server.ex
+++ b/lib/kafka_ex/server.ex
@@ -582,11 +582,7 @@ defmodule KafkaEx.Server do
             state.api_versions
           )
 
-        updated_state = %{
-          state
-          | metadata: metadata,
-            correlation_id: correlation_id
-        }
+        updated_state = %{state | correlation_id: correlation_id}
 
         {:reply, metadata, updated_state}
       end


richardlarocque:
I tried to run the tests on this PR locally, as described in CONTRIBUTING.md.  They didn't quite work, but they didn't work for me on `master` either, so this does not seem to be a regression specific to this PR.

----

The idea behind this PR is as follows.

When we try to produce a message, the client will end up having to update its metadata if it doesn't currently have metadata for the given topic (https://github.com/kafkaex/kafka_ex/blob/92557042cbdc971f482eb3b540db6f2dc13b06e3/lib/kafka_ex/server.ex#L437-L459).  That logic seems to assume (reasonably) that the contents of the `metadata` state variable are an all-topic metadata response.

However, an explicit request for topic-specific metadata could, prior to the change presented here, replace the all-topic metadata maintained by the periodic timer with a topic-specific metadata response.  That would be bad because the next produce request would then be more likely to have to go re-fetch all-topic metadata.

That would be inefficient.  Prior to https://github.com/kafkaex/kafka_ex/pull/395, this bug would result in sending a lot of traffic at the "lead" broker, potentially overwhelming it with all-topic metadata requests.  I suspect that PR partitally mitigates the effect since the metadata requests would be more evenly spread out and you could horizontally-scale your way past the problem, but still, the inefficiency would remain.

jbruggem:
@joshuawscott what's our policy on approvals ? it seems that 2 is enough for such a simple change.

joshuawscott:
oh yeah, 2 is plenty.

yingyingtang-brex:
do we need to update the other server implementation with this change, e.g. [here](https://github.com/kafkaex/kafka_ex/blob/9e0dca7ffe81c47116963005ae6c5bf0c45c228f/lib/kafka_ex/server_0_p_10_and_later.ex#L166)?

PR #413, opened 2020-08-04
Add `ssl_options` to consumers

markusfeyh:
This allows the pass-through of `ssl_options` when starting workers starting a consumer or consumer group.


markusfeyh:
Failing tests from `7181920` are unrelated and passed locally as well as when re-running the build

markusfeyh:
Giving this a bump since it's pretty minor, could you take a look?
@joshuawscott @jbruggem

test/integration/consumer_group_test.exs
```diff
@@ -24,6 +24,26 @@ defmodule KafkaEx.ConsumerGroup.Test do
     assert consumer_group == :no_consumer_group
   end
 
+  test "create_worker allows us to pass in use_ssl and ssl_options options" do
+    Application.put_env(:kafka_ex, :use_ssl, true)
```
dantswain, line 28: You'd need to reset this to the previous value at the end of your test.

test/integration/consumer_group_test.exs
```diff
@@ -24,6 +24,26 @@ defmodule KafkaEx.ConsumerGroup.Test do
...
+    ssl_options = Application.get_env(:kafka_ex, :ssl_options)
+    assert ssl_options == Config.ssl_options()
+
+    custom_ssl_options = [
+      cacertfile: File.cwd!() <> "/ssl/ca-cert-custom",
+      certfile: File.cwd!() <> "/ssl/cert-custom.pem",
+      keyfile: File.cwd!() <> "/ssl/key-custom.pem"
```
dantswain, line 35: Can we just reuse the files we already have?  I'm also not sure how the test would pass if we don't configure the test broker to use these.

markusfeyh: Yah, these are the same as the existing files we already have, just the filename is different so it can be validated in the test that a different file is being used.

dantswain: Ahh that makes sense.  It would be nice if we could document that somehow (even if it's just in a comment in this test).  There's about a 100% chance that some day someone will wonder why we have a duplicate set of keys with different names and get very confused.  Probably me because I'll forget :joy:.  Another option might be to create a copy of the files as part of the test run and then delete them during clean up.

test/integration/consumer_group_test.exs
markusfeyh: Sure, will do

test/integration/consumer_group_test.exs
markusfeyh: Added some documentation, and realized it would be even more explicit to use a symbolic file links to prevent duplicate files combined with a comment. Would that work?

ssl/ca-cert-custom
```diff
@@ -0,0 +1 @@
+ca-cert
```
markusfeyh, line 1: These are symbolic links (soft links)

test/integration/consumer_group_test.exs
markusfeyh: On second though, I also really like your proposal @dantswain of creating and deleting the files to not confuse the repo since others may look at those certs for examples of how to configure client.

markusfeyh:
Could you give it another round of review @dantswain ?

test/integration/server0_p_10_and_later_test.exs
```diff
@@ -24,12 +24,14 @@ defmodule KafkaEx.Server0P10P1AndLater.Test do
       Enum.member?(existing_topics(), name)
     end)
 
-    assert @num_partitions ==
-             KafkaEx.Protocol.Metadata.Response.partitions_for_topic(
-               KafkaEx.metadata(),
-               name
-             )
-             |> Enum.count()
+    wait_for(fn ->
```
markusfeyh, line 27: This will explicitly raise an error if it the comparison does not return `true` which fails the test

```
  1) test can create a topic (KafkaEx.Server0P10P1AndLater.Test)
     test/integration/server0_p_10_and_later_test.exs:9
     ** (RuntimeError) too many tries waiting for condition
     code: wait_for(fn ->
     stacktrace:
       test/test_helper.exs:165: TestHelper.wait_for_value/5
       test/integration/server0_p_10_and_later_test.exs:27: (test)
```

markusfeyh:
Giving this a bump going into next week, could you take a look?

markusfeyh:
I saw @joshuawscott provided an option to specify the worker and then pass that in as part of the fetch_options. The downside of that approach is that we would lose the benefit of the `ConsumerGroup.Manager` supervision tree. 

The impact of that is that we would have to maintain more code around using the library, as a pose to this PR which would allow different SSL options to be passed through as with other consumers in the library. 

I'm trying to think of a strong reason why this is a bad idea, and see mostly positives. Could you provide feedback?

joshuawscott:
I think the SSL options do more properly go into the worker rather than fetch options, since there are other interactions (metadata, commit) that aren't fetch.

Argonus:
Hi @dantswain / @joshuawscott 
Any news when this can be merged?

dantswain (approved):

Issue #397, opened 2020-02-19
Add kafka headers to messages, producer and consumer


matreyes:
Kafka protocol >= 0.11 allows to send arbitrary headers [{key, value}] along with the message payload. This is very useful for metadata as schema version, origin, etc.

habutre:
I am interested in work on this, please let me know how can I help.
It also involves the `kayrock` client, right?

jbruggem:
@habutre PRs are welcome of course :-). What kind of information do you need to get started ? Some pointers in the code to know where to get started ?

habutre:
The main point would be if supporting headers should be first implemented on Kayrock and back ported to kafka_ex or the other way around.  

I will check the points in the code where changes will be done but if there is a well-know points where this change will affects would be handy

dantswain:
Kayrock already has support for headers in `Kayrock.RecordBatch.Record` - https://github.com/dantswain/kayrock/blob/master/lib/kayrock/record_batch.ex

It looks like it’s expecting headers to be a binary value and that gets serialized directly. I can do a little digging later today.

dantswain:
OK I was half-wrong.  Kayrock has the headers field for the v2 message format but it does not appear to properly serialize/deserialize the values.  So you'd need to support that in Kayrock first. 

The message spec for headers is here: https://kafka.apache.org/documentation/#recordheader

If you need help with that, feel free to reach out.  I've been quite busy but I can find time.

habutre:
Hey @dantswain  in order to validate the  support for record headers on `Kayrock` I have pushed a small branch with a test case and a temporary change on deps to point `KafkaEx` to `Kayrock` master

I don't know exactly what would be the best approach here to validate Kayrock's  integration and if that single test case is enough to validation purposes.

I have checked other integration tests and couldn't find any similar test using `RecordBatch.Record` 

Locally all tests passes, sometimes in the second or third run.

yingyingtang-brex:
Hi! As we have [the above fix](https://github.com/kafkaex/kafka_ex/pull/414) fixed merged on Nov 8, 2021, is there a plan to have a new release including this fix?

habutre:
@joshuawscott  according to the [latest release](https://github.com/kafkaex/kafka_ex/releases/tag/0.13.0)  this issue can be closed, right?

PR #414, opened 2020-08-13
Kafka 11 - record headers support

habutre:
This PR is targeted to #397 to provide support for record headers provided since Kafka 0.11

The headers were implemented first on `Kayrock` and for testing purposes this branch points to Kayrock master. After a new  Kayrock be released this PR can be changed and merged. This change is only for your review and validation.

To maintain compability while using the new client the `Protocol.{Produce, Fetch}.Message`  had to be changed.

Only the new client is aware about such info on messages and may not cause conflicts or misbehavior

Out of scope: The mix format causes some changes that should be evaluate if make sense or have to be reverted


habutre:
Hey @dantswain @jbruggem @joshuawscott @bjhaid I think you are the right guys to involve here to help me (let me know if someone might be added or removed from my mentions)


I am facing random errors with `ci_tests.sh` also with `all_tests.sh` locally, did you have some tips & tricks that I can apply to get rid of it and making the tests pass?

One can see in the CI tests each config has erros in different test cases and I am a bit lost to find a solution.

Thanks for any help.

joshuawscott:
When I released 0.11.0 I fixed most of the intermittent failures, I thought. I was able to run 20x in a row without any failures locally. That said, because Travis uses 1.5 core machines, it doesn't necessarily work the same. There does seem to be a completely consistent failure, though:
```
  1) test gzip compression - produce v3, fetch v0 (KafkaEx.KayrockRecordBatchTest)
     test/integration/kayrock/record_batch_test.exs:301
     ** (UndefinedFunctionError) function nil.value/0 is undefined
     code: assert message.value == msg
     stacktrace:
       nil.value()
       test/integration/kayrock/record_batch_test.exs:328: (test)
     The following output was logged:
     
     21:27:44.845 [debug] Successfully connected to broker "localhost":9093
     
     21:27:44.888 [debug] Successfully connected to broker "localhost":9094
     
     21:27:44.932 [debug] Successfully connected to broker "localhost":9095
```
Often a single failure will cause ripples due to using the same underlying brokers, so my suggestion would be to solve this failure and see if the others begin to work.

habutre:
Hey @dantswain  I would really appreciate if you could take a look into my last commit that make the ci tests pass. 
I am a bit affraid to change pre-existent tests to make all them pass. 

@joshuawscott  I think now you can review and if you can help to clarify about the offsets manipulation on ` kafka_ex/test/integration/kayrock/record_batch_test.exs` will be awesome

:information_source:  Just a remind I am pointing the `Kayrock` to master and a release there is necessary before move forward with this PR

joshuawscott:
kayrock 0.1.12 is out now

dantswain:
@habutre I apologize for not getting to this yet.  I have been unbelievably busy :(  I will try to make time.

lib/kafka_ex/protocol/fetch.ex
```diff
@@ -66,6 +67,7 @@ defmodule KafkaEx.Protocol.Fetch do
             offset: integer,
             key: binary,
             value: binary,
+            headers: list(),
```
dantswain, line 70: We should be able to be more specific here, I think?  A list of what?

dantswain (commented):
@habutre I think this all looks good in theory, but it's hard to be sure of what I'm reviewing with all of the `mix format` changes included.  Maybe after #420 is merged you could merge  into your branch and update this PR?

lib/kafka_ex/protocol/fetch.ex
habutre: I think now is clearer

habutre:
Is it possible to retry the `CI`? Checking the errors 1) timeout 2) failed to get deps

test/integration/kayrock/record_batch_test.exs
```diff
@@ -234,7 +287,7 @@ defmodule KafkaEx.KayrockRecordBatchTest do
 
     fetch_responses =
       KafkaEx.fetch(topic, 0,
-        offset: max(offset - 2, 0),
+        offset: offset,
```
dantswain, line 290: I don't remember exactly why this was set this way - probably some kind of bizarre test behavior.  Did you have a compelling reason to change it?

dantswain (commented):
I think we're good here.  I kicked the build again and had one question about what seems like an unrelated change.

test/integration/kayrock/record_batch_test.exs
habutre: At first I tried to understand the reason to get a previous offset instead the one returned by the producer, but since it always worked before I kept is AS-IS. After I have introduced the headers the kayrock tests start failing randomly but always with the same reason `nil` message.

I watch the topic with Kafka tool and them I figure out that the offset - 2 doesn't match with the expected message then I tried to remove the `max(offset - 2, 0)` and then I got success on every execution what make things that is the reason and any other test failed

I don't have a rational about that behavior only that on my mind the offset should not be manipulated to get the desired message, it was just a recursively try-n-check.

Argonus:
@habutre @joshuawscott 
Are we blocked by sth here? Maybe it's worth merging master to this PR & if everything is green merge it to the main branch.
If I can help somehow, please let me know.

habutre:
> @habutre @joshuawscott
> Are we blocked by sth here? Maybe it's worth merging master to this PR & if everything is green merge it to the main branch.
> If I can help somehow, please let me know.

Hey @Argonus  this PR is ready for review and I am keeping as much as possible synced with latest merged PRs.

habutre:
> @habutre @joshuawscott
> Are we blocked by sth here? Maybe it's worth merging master to this PR & if everything is green merge it to the main branch.
> If I can help somehow, please let me know.

Hey @Argonus  this PR is ready for review and I am keeping as much as possible synced with latest merged PRs.

test/integration/kayrock/record_batch_test.exs
jbruggem: @dantswain in case you're interested, here is the commit where you added this `max(offset - 2, 0)`:  7197a909922ff5a2f75de4ee445ea9b9099395ac

I didn't see anything in particular to help me understand. IMO the new changed code makes more sense :shrug:

jbruggem (approved):
LGTM. I'd rather have a second opinion though :)

lib/kafka_ex/protocol/produce.ex
```diff
@@ -53,8 +53,14 @@ defmodule KafkaEx.Protocol.Produce do
...
     - timestamp: timestamp (`kafka_version: "kayrock"` ONLY)
     """
-    defstruct key: nil, value: nil, timestamp: nil
-    @type t :: %Message{key: binary, value: binary, timestamp: integer}
+    defstruct key: nil, value: nil, headers: nil, timestamp: nil
+
+    @type t :: %Message{
+            key: binary,
+            value: binary,
+            headers: list(),
```
jbruggem, line 61: should this be the same as in Fetch ?

```suggestion
            headers: [{key::binary, value::binary}],
```

jbruggem (approved):
Still LGTM. @joshuawscott @dantswain could one of you cross-check these changes to make sure it makes sense ? I'd rather not approve alone.

dantswain (approved):
Yeesh sorry for the delay here.  @habutre can you check @jbruggem 's comment, but otherwise looks good to me.  Thanks for the work on this and for your patience!

lib/kafka_ex/protocol/produce.ex
habutre: @jbruggem good catch it should be the same

lib/kafka_ex/protocol/produce.ex
```diff
@@ -53,8 +53,14 @@ defmodule KafkaEx.Protocol.Produce do
     - value: is the message to be written to Kafka logs.
```
habutre, line 53: What about header's doc?

```suggestion
    - value: is the message to be written to Kafka logs.
    - headers: is a list of keys and values that adds metadata to messages
```

jbruggem: Indeed

jbruggem:
@habutre do you wish to commit those last documentation suggestions ?

Once done, I'll merge.

Argonus:
@jbruggem Maybe it's worth releasing a new version as well? 
There would be two quite big things: support for snappy & support for headers

jbruggem:
@Argonus TBH I've never cut a release on KafkaEx before, so I'm not sure what the criteria are :). 

@dantswain @joshuawscott would this be a good moment to make a release, and if so who can do it ?

PR #420, opened 2020-11-02
Improve docker-compose by adding healthcheck condition at startup

jbruggem:
This adds a health check and a dummy service to make sure that we wait for Kafka to be healthy when we start the nodes.

This also adds an environment variable to toggle the capture of logs in tests, which is convenient when debugging complicated test cases. It also adds a formatter to show module, function and PID.

There's some changes due to a `mix format`, but other than that this MR does not change any elixir code.


dantswain:
Seems like it's not working :(  I like the idea, though.

jbruggem:
> Seems like it's not working :( I like the idea, though.

The two failures seem to be:
- a coverall result upload error
- tests failing in *one of the pipelines*, but I didn't change any elixir code ? (mix format did, but I trust that :P)

Could it just be pipeline unreliability ?

I can't seem to re-trigger the failed jobs. Would it be a good idea to do that ?

dantswain:
Triggered again.

I have seen the coveralls error before and don't have a good solution.  I sort of wonder if it's worth keeping coveralls at this point... i like having coverage metrics but it's not like we ever do anything about it.  We could keep the coveralls lib installed because it does produce a nice report, but disable the travis integration.  That would be for a different PR anyways.

You could also try doing the `mix format` changes on a separate pr just to be safe?  Let's see what the build does.

dantswain (approved):

jbruggem:
> I have seen the coveralls error before and don't have a good solution. I sort of wonder if it's worth keeping coveralls at this point... i like having coverage metrics but it's not like we ever do anything about it. We could keep the coveralls lib installed because it does produce a nice report, but disable the travis integration. That would be for a different PR anyways.

It failed only on that this time. I'll disable the travis integration here if you don't mind ?

jbruggem:
@dantswain OK for you if I merge with the coveralls deactivated ?

dantswain:
Can you maybe re-enable it and then open a separate PR?  I think it's probably the right move but I think it deserves its own review and weigh in from others.

jbruggem:
> Can you maybe re-enable it and then open a separate PR? I think it's probably the right move but I think it deserves its own review and weigh in from others.

OK. I don't think we'll manage to get a green pipeline though :)

dantswain:
> OK. I don't think we'll manage to get a green pipeline though :)

That's fine, we know why.

jbruggem:
Given the low impact, I'll merge this then, since it seems it could help review #414

Issue #421, opened 2020-11-05
Investigate alternatives to travis-ci


dantswain:
Travis-CI seems to be getting noticeably slower for OSS builds: https://www.traviscistatus.com/#month

I've definitely noticed significant queue times for builds, which is a pretty major slowdown for iterating on PRs.  Furthermore, using travis has caused some difficulty in the test suite because the instances it runs are somewhat anemic and that can make running the tests and kafka at the same time fairly rough.

Alternatives that I have a some amount of familiarity with off-hand:
* Github actions
* CircleCI
* GitLab (can we do this without moving the project?)

I'm open to opinions here, though I'll be favoring platforms that the core maintainers are comfortable with.  @joshuawscott @jbruggem @bjhaid

jbruggem:
Excellent initiative !

I have a tiny bit of experience with Github actions, and a lot of experience with Gitlab CI (that what we use at my company).

My opinion is that we should consider first Github actions. There's a strong added value in having everything in one tool ! My experience with it was good.


If it doesn't pan out, I can voutch for GitlabCI (though I have experience with it on-prem, so I don't know the performance it has for free accounts on their cloud version).  Not sure about how we could integrate the CI status in Github though :thinking: . In that case, I'd recommend using their `services` system rather than a docker-compose, which will be more efficient. We'd keep the docker-compose setup locally.

dantswain:
I gave github actions a try for kayrock and it was pretty easy and builds pretty fast.  https://github.com/dantswain/kayrock/blob/master/.github/workflows/elixir.yml

The hard part will be kafka of course

PR #431, opened 2021-01-07
Add Github actions CI

gerbal:
Prompted by #421 and #422 - I took a whack at writing some github actions to run the test suite and added some additional checks. 

I've split the integration tests apart to run in a matrix. I was hoping this would decrease the flakeyness in testing, but if anything it seems a bit worse. On the plus side, it's a bit faster to run all the tests in parallel tasks. Are there tweaks to the kafka configuration to let it run better in small vms? 

I have also added common CI code quality checks. Currently this includes `credo`, `mix format --check-formatted`, `mix compile --warnings-as-errors --force`, and `dialyzer`.


dantswain:
@gerbal  Thanks so much for taking this on!  It would be a huge help if you could get this to work reliably!

Generally what seems to happen with the integration tests is that because we need to run a 3-node kafka cluster AND the tests, any cpu-constrained environment will have all kinds of bad behavior because that resource is constrained.  I think we want to have at least 2 nodes (3 is probably better) in our test cluster because otherwise we aren't going to exercise any of the tricky metadata juggling.

gerbal:
Stray misclick

gerbal:
Thanks @dantswain! I'll keep hacking on this as I have time

PR #437, opened 2021-02-23
Support snappyer 2

Argonus:
As the original [PR](https://github.com/kafkaex/kafka_ex/pull/391) looks a bit abandoned I've created new one cherry-picks of source and minor fixes.

- Resolved conflicts
- Use `snappy-erlang-nif` as default snappy compression 
- Add `snappyer` as optional snappy compresions

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -2,7 +2,7 @@ use Mix.Config
 
 config :ex_unit, capture_log: is_nil(System.get_env("SHOW_LOGS"))
 
-config :kafka_ex, snappy_module: :snappyer
+config :kafka_ex, snappy_module: :snappy
 
 config :kafka_ex, sync_timeout: 60_000
 
--- a/config/test.exs
+++ b/config/test.exs
@@ -2,8 +2,9 @@ use Mix.Config
 
 config :ex_unit, capture_log: is_nil(System.get_env("SHOW_LOGS"))
 
-config :kafka_ex,
-  sync_timeout: 60_000
+config :kafka_ex, snappy_module: :snappyer
+
+config :kafka_ex, sync_timeout: 60_000
 
 # Help debug tests that are tricky to understand
 config :logger, :console,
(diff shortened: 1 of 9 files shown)


lib/kafka_ex/compression.ex
```diff
@@ -58,7 +58,11 @@ defmodule KafkaEx.Compression do
         <<valsize::32-unsigned, value::size(valsize)-binary, rest::binary>>,
         so_far
       ) do
-    {:ok, decompressed_value} = :snappy.decompress(value)
+    {:ok, decompressed_value} = snappy_module().decompress(value)
     snappy_decompress_chunk(rest, so_far <> decompressed_value)
   end
+
+  defp snappy_module do
+    Application.get_env(:kafka_ex, :snappy_module, :snappyer)
```
Argonus, line 66: Wrong default.

config/test.exs
```diff
@@ -2,8 +2,9 @@ use Mix.Config
 
 config :ex_unit, capture_log: is_nil(System.get_env("SHOW_LOGS"))
 
-config :kafka_ex,
-  sync_timeout: 60_000
+config :kafka_ex, snappy_module: :snappyer
```
Argonus, line 5: Wrong default

Argonus:
@joshuawscott Hi
Kayrock related PR is here: https://github.com/dantswain/kayrock/pull/19
I'm also thinking about making it a bit smarter and in the case of using "kayrock" instead of using snappy use kayrock snappy module.  What do you think about it? (I don't really get relationship between this two libs ;P)

jbruggem (approved):

Argonus:
@joshuawscott I've no power to restart travis checks (even on my PR...) but they look like a random fail. Especially that one with timeout on docker pull. If you could be so kind and restart 🙏

joshuawscott (approved):
:shipit:

PR #446, opened 2021-10-28
Update metadata before topic creation

jbruggem:
We have infrequently gotten in production the following errors while trying to create a topic:

```
(MatchError) no match of right hand side value: {:error, :not_controller}
```

From reading the code of KafkaEx, it seems to me that this could possibly happen if we haven't updated the metadata in a while. It's hard to confirm, but it seems coherent with what I've read from the code and what we experienced.

In this PR, I propose to systematically update the metadata before selecting the controller broker to create the topic.

Additionally, for older versions of Kafka we didn't update our metadata in the state when we fetched metadata for the user in `kafka_server_metadata`. I changed that.

### DIFF
--- a/lib/kafka_ex/server_0_p_10_and_later.ex
+++ b/lib/kafka_ex/server_0_p_10_and_later.ex
@@ -268,13 +268,17 @@ defmodule KafkaEx.Server0P10AndLater do
     {:reply, response, state}
   end
 
-  defp request_to_controller(main_request, state) do
+  defp request_to_controller(main_request, stale_state) do
+    # We need to make sure that the metadata is up to date,
+    # to maximize the chance that the controller hasn't switched
+    # to another broker.
+    state = update_metadata(stale_state)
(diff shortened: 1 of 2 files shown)


lib/kafka_ex/server_0_p_10_and_later.ex
```diff
@@ -268,13 +245,17 @@ defmodule KafkaEx.Server0P10AndLater do
     {:reply, response, state}
   end
 
-  defp request_to_controller(main_request, state) do
+  defp request_to_controller(main_request, stale_state) do
+    # We need to make sure that the metadata is up to date,
+    # to maximize the chance that the controller hasn't switched
+    # to another broker.
+    state = update_metadata(stale_state)
```
dantswain, line 252: I was a little concerned about this impacting performance but it looks like it's only called on create/delete topics, so :+1:

dantswain (approved):
Thanks @jbruggem !

PR #450, opened 2022-02-09
Use offset_fetch api_version when fetching messages

GDegrove:
When switching to the new kayrock client, we noticed that the api_version for the auto_commit of the fetch message is not taken into account.
Therefore, even if using the API version 2 for the offset commit with kayrock, the next fetch will always use the api_version 0

This introduces a divergence in the way the new client can be used compared to the old implementation. As it's up to the user code to provide the correct offset to fetch.

### DIFF
--- a/lib/kafka_ex.ex
+++ b/lib/kafka_ex.ex
@@ -297,9 +297,21 @@ defmodule KafkaEx do
     api_version = Keyword.get(opts, :api_version, 0)
     # same for offset_commit_api_version
     offset_commit_api_version = Keyword.get(opts, :offset_commit_api_version, 0)
+    # By default, it makes sense to synchronize the API version of the offset_commit and the offset_fetch
+    # operations, otherwise we might commit the offsets in zookeeper and read them from Kafka, meaning
+    # that the value would be incorrect.
+    offset_fetch_api_version = Keyword.get(opts, :offset_fetch_api_version, offset_commit_api_version)
+
 
     retrieved_offset =
-      current_offset(supplied_offset, partition, topic, worker_name)
+      current_offset(
+        supplied_offset,
+        partition,
+        topic,
+        worker_name,
+        offset_fetch_api_version,
+        nil
+      )
 
     Server.call(
       worker_name,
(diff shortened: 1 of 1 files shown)


lib/kafka_ex.ex
```diff
@@ -299,7 +299,14 @@ defmodule KafkaEx do
     offset_commit_api_version = Keyword.get(opts, :offset_commit_api_version, 0)
 
     retrieved_offset =
-      current_offset(supplied_offset, partition, topic, worker_name)
+      current_offset(
+        supplied_offset,
+        partition,
+        topic,
+        worker_name,
+        offset_commit_api_version,
```
jbruggem, line 307: Nit-pick on this: I'm wondering if we shouldn't offer a choice to the user whether they wish to use `offset_commit_api_version` also for `offset_fetch_api_version`.

In any case, I'm thinking that maybe some explicit variable name would help the reader follow the logic and make it explicit that we've decided that `offset_fetch_api_version = offset_commit_api_version`.

Suggestion:

```elixir
    offset_commit_api_version = Keyword.get(opts, :offset_commit_api_version, 0)
    # By default, it makes sense to synchronize the API version of the offset_commit and the offset_fetch
    # operations, otherwise we might commit the offsets in zookeeper and read them from Kafka, meaning
    # that the value would be incorrect.    
    offset_fetch_api_version = Keyword.get(opts, :offset_fetch_api_version, offset_commit_api_version)

    retrieved_offset =
      current_offset(
        supplied_offset,
        partition,
        topic,
        worker_name,
        offset_fetch_api_version,
```

jbruggem (commented):
Apart from my comment, this looks good to me. Given that it's a sensitive piece of code and that I worked on this with you, I'd rather we had a second approval.

@dantswain @joshuawscott could either of you take a look ? It's a short one :)

lib/kafka_ex.ex
GDegrove: It makes real sense, I pushed the correction :)

jbruggem (approved):
OK for me. Once we have a second approval, we can merge.

bjhaid (approved):
lgtm

jbruggem:
Thanks a lot @bjhaid !

PR #457, opened 2022-03-25
FIX: Partition assignment when topic doesn't exist

Zarathustra2:
Fixes an issue when having a key and the topic does not exist.
Previously this would cause the genserver to terminate:

mfa=:gen_server.error_info/7 [error] GenServer :greeks_worker_4 terminating
** (ArithmeticError) bad argument in arithmetic expression
    :erlang.rem(2111275913, 0)
    (kafka_ex 0.12.1) lib/workers/greeks/greek_server.ex:418: KafkaEx.DefaultPartitioner.assign_partition_with_key/3
    (kafka_ex 0.12.1) lib/kafka_ex/server.ex:400: KafkaEx.Server0P10AndLater.kafka_server_produce/2
    (stdlib 3.17) gen_server.erl:721: :gen_server.try_handle_call/4
    (stdlib 3.17) gen_server.erl:750: :gen_server.handle_msg/6
    (stdlib 3.17) proc_lib.erl:226: :proc_lib.init_p_do_apply/3

### DIFF
--- a/lib/kafka_ex/default_partitioner.ex
+++ b/lib/kafka_ex/default_partitioner.ex
@@ -70,10 +70,19 @@ defmodule KafkaEx.DefaultPartitioner do
        ) do
     hash = Murmur.umurmur2(key)
 
-    partitions_count =
-      metadata |> MetadataResponse.partitions_for_topic(topic) |> length()
+    partitions = metadata |> MetadataResponse.partitions_for_topic(topic)
 
-    partition_id = rem(hash, partitions_count)
+    partition_id =
+      case partitions do
+        # Topic doesn't exist yet, no partitions
+        [] ->
+          0
+
+        xs ->
+          partitions_count = xs |> length()
+          rem(hash, partitions_count)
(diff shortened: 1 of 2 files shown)


Zarathustra2:
Actually I think, this is bad idea. A future message with the same key could be going into a different partition if the auto created topic has more than 1 partition. I am not sure what the best approach would be in that case

Argonus:
@Zarathustra2 We've had a similar issue. Instead of assigning a consumer to nonexisting data, we've just added health checks that observe that the consumer is connected to all topics & partitions it should. If it's not, we are restarting it a few times & killing the runtime after some time.

dantswain:
Hi @Argonus and @Zarathustra2 !

I think maybe the best thing here is to raise an error if this condition is encountered?  Attempting to publish to a topic that doesn't exist is kind of undefined behavior to begin with, so maybe we just give the user a more helpful error message so that they can decide how to resolve it for their use case?

PR #460, opened 2022-04-11
GitHub actions checks cdn

Argonus:


Argonus:
@dantswain @joshuawscott 
Done, after some long fight I've given up and just use the same approach as kafkajs is doing, so moved to confluent-based images.
Currently, there is only the newest version of elixir & erlang in the matrix, but I'm happy to play with other ones as well. What is earliest version we have to support?

dantswain:
@Argonus This is amazing!

The earliest version of Elixir we support is 1.6.  If changing that makes this library easier to maintain then I'm totally in favor.

I hate to ask this after all the work you put in, but it looks like there's a bunch of formatting changes here... would you mind making those on a separate PR to make it easier to understand what changed in order to support github actions?  Thanks!

Argonus:
Sure thing, I propose to let's make static checks work only for ~newst version of elixir (that will support 1.6 syntaxes) & run tests for 1.6, 1.10 & 1.13  & corresponding OTP.

dantswain:
Sounds good!

Argonus:
Versions added.
Static code is currently locked at Elixir Version 12, to bump it to version 13 I'll have to drop support for Elixir 1.6.
It looks like sth that could be done in the next major release.

Argonus:
@dantswain Just a reminder ;)

test/integration/server0_p_10_p_1_test.exs
```diff
@@ -11,40 +11,54 @@ defmodule KafkaEx.Server0P10P1.Test do
     # note this checks against the version of broker we're running in test
     # api_key, max_version, min_version
     api_versions_kafka_0_11_0_1 = [
-      [0, 3, 0],
```
dantswain, line 14: Are we testing against a newer version of Kafka now?  If so, can we update the README (IIRC it states what version we test against) and the variable names here (`kafka_0_11_0_1` etc)?

test/socket_test.exs
```diff
@@ -69,96 +75,112 @@ defmodule KafkaEx.Socket.Test do
...
-    KafkaEx.Socket.close(socket)
-  end
+  describe "with ssl socket" do
+    setup do
+      SSLServer.start(3030)
+      {:ok, [ssl_port: 3030]}
+    end
+
+    test "create a SSL socket", context do
+      {:ok, socket} =
```
dantswain, line 145: Huh it looks like what happened here is the test failed but then it succeeded on the second run.  I wonder if there's a way to clear these when something like that happens.  I'm not 100% that's even the right thing to do but I'm curious if that's an option.

Argonus: I'm not sure. The socket was closed too fast, but it never happened to me in the local setup.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,108 @@
...
+        run: mix compile --force
+
+      - name: Check Code Format
+        run: mix format --check-formatted
+
+      - name: Run Credo
+        run: mix credo
+
+      - name: Run Dialyzer
+        run: mix dialyzer --no-check
```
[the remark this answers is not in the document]
Argonus: As we are always building a plt in the step above, (or checking against the currently existing one) there is no need to double-check it here. So I'm skipping that part :) 
It gives some performance.

PR #464, opened 2022-05-04
Apply mix format

Argonus:
Apply mix format to update code quality before enabling github actions checks.


Argonus:
@dantswain Once this will be merged, I'll bump https://github.com/kafkaex/kafka_ex/pull/460 that branch, then all should be clean there ;)

PR #467, opened 2022-06-02
Migrate GitHub Actions to erlef/setup-beam

kianmeng:
See https://github.com/erlef/setup-beam/issues/20

Also fixed the typos.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -27,8 +27,8 @@ jobs:
         - name: Checkout Github repo
           uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


Argonus:
Shhh... my bad... I've not caught that case with the cache invalidation. PLT will fail once the minor OTP version has been updated. 
Probably removing `--no-check` or adding minors to the version will fix it. I'd just remove `--no-check` it gives a little performance boost, but it shouldn't be an issue here. 

@kianmeng Could you fix it here?

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #469, opened 2022-06-08
Use full versions for OTP & Elixir in static checks

Argonus:
As we are using caching to build PLT, and cache key relay on OTP & Elixir version we have to use full versions there.


jbruggem (approved):
Thanks !

jbruggem:
Thanks !

PR #470, opened 2022-07-27
Added examples for topic create/delete functions.

alfetahe:
Added examples in the topic create/delete function doc comments.


jbruggem (approved):
Thanks for the contribution !

Issue #390, opened 2019-12-24
GenConsumer supervisor necessarily launches consumer module as a worker


tanguilp:
In https://github.com/kafkaex/kafka_ex/blob/master/lib/kafka_ex/gen_consumer/supervisor.ex#L92, we can see that the child is launched as a worker.

As a consequence, the shutdown timeout is automatically set to 5_000 ms (among other default settings). However, the child could be implemented as a supervisor, whose timeout would be `:infinity` (the supervisor being in charge of handling his children termination in time).

I suggest using the child specs mechanism and rewriting it to:

```elixir
{gen_consumer_module, [consumer_module, group_name]}
```

so that `child_spec/1` is called.

What do you thing?

joshuawscott:
I think refactoring to use the child specs would be good, and I think will be required to support Elixir 1.10 without generating compile warnings.

KafkaEx.GenConsumer does additional work besides just supervising, so making it a Supervisor might be problematic.

PR #473, opened 2022-08-30
Fix compiler warnings

b1az:
The usage of `child_spec/1` addresses #390.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ KafkaEx
 [![API Docs](https://img.shields.io/badge/api-docs-yellow.svg?style=flat)](http://hexdocs.pm/kafka_ex/)
 
 KafkaEx is an Elixir client for [Apache Kafka](http://kafka.apache.org/) with
-support for Kafka versions 0.8.0 and newer.  KafkaEx requires Elixir 1.5+ and
+support for Kafka versions 0.8.0 and newer. KafkaEx requires Elixir 1.6+ and
 Erlang OTP 19+.
 
 See [http://hexdocs.pm/kafka_ex/](http://hexdocs.pm/kafka_ex/) for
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -10,15 +10,15 @@ KafkaEx
 [![API Docs](https://img.shields.io/badge/api-docs-yellow.svg?style=flat)](http://hexdocs.pm/kafka_ex/)
 
 KafkaEx is an Elixir client for [Apache Kafka](http://kafka.apache.org/) with
-support for Kafka versions 0.8.0 and newer.  KafkaEx requires Elixir 1.5+ and
+support for Kafka versions 0.8.0 and newer. KafkaEx requires Elixir 1.6+ and
```
joshuawscott, line 13: I believe `Config` as a replacement for `Mix.Config` was introduced in 1.9. We would need to update this to say 1.9 if we make that change. We should also change mix.exs to require the same version.

b1az: This proposed version bump to `1.6+` (e513413) was meant to get README in sync with [the already-changed mix.exs](https://github.com/kafkaex/kafka_ex/commit/f2870c8d8566448f97b1b0493b7e642c3191f270#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308R9). 

While the migration to `Config` module (1f4e99a), [introduced in 1.9.0](https://github.com/elixir-lang/elixir/blob/v1.9/CHANGELOG.md#elixir-2), was meant to remove the warning for KafkaEx developers. This change has no bearing on KafkaEx users. Though it was my oversight that Elixir 1.6 was still used in the test suite.

Hence I think it's best if I drop 1f4e99a. A decision to bump versions from 1.6 to 1.9 can be made at some other time.

joshuawscott: That makes sense to me 👍

Argonus: @joshuawscott Can we merge this?

joshuawscott (approved):

Issue #434, opened 2021-02-09
How to gracefully shutdown consumer group?


lessless:
Hi,

I'm looking for a way to gracefully shutdown a  consumer group. 

Let's say there is a setting that can be flipped to turn off consumption and that setting can be checked in `handle_message_set`. 

```
 def handle_message_set(message_set, state) do
    do_some_job(message_set)
    if Settings.stop?() do
      commit_and_shutdown()
    else 
      commit()
    end
  end

```

The absolute must here is to no receive any new messages even if the cluster undergoes a rebalance.

PR #474, opened 2022-08-30
Enable graceful shutdown of a GenConsumer

b1az:
Add a user-configurable `:shutdown` to `GenConsumer`'s child_spec.

Addresses #434, if `do_some_job`, from the referred issue, is either synchronous or async. Not sure if the change covers the case of a cluster rebalance.


b1az:
@Argonus and @joshuawscott, I've been already using the proposed change, as the worker's (default) `:shutdown` value of 5 s is, in my use, not long enough to let all messages in a `message_set` be processed in time. Even though the proposed change is likely the smallest possible/needed to achieve a graceful shutdown, I might have missed something. 
 
If you'll find time, any feedback, either affirmative or dismissive, would be much appreciated. (Especially given that I already went ahead and use this in production.)

Argonus:
@joshuawscott I've tested that on my own setup, it works well with different shutdowns time.

PR #475, opened 2023-02-19
[Kayrock Part 0] Add healthcheck for zookeepr

Argonus:
I've found issue once zookeeper is unhealthy, errors is raised as unhealthy kafka. Here, it will fail on zookeeper level.


dantswain (approved):
Thanks!!

Argonus:
@joshuawscott can we merge this?

PR #478, opened 2023-02-22
[Kayrock Part 0] Bump elixir to 1.8

Argonus:
## Overview
The current elixir version is 1.14. Soon probably, 1.15 will be released. There are many breaking changes, deprecation warnings & compile issues between different versions. On our road to `v1.0.0`. We will have to fight with all of them. 
Let's bump the version to elixir v1.8.x which will allow us to solve deps & compile warnings, then we can bump higher if we want to.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,12 +15,14 @@ jobs:
     strategy:
       matrix:
         pair:
+          - elixir: 1.14
+            otp: 25.2
           - elixir: 1.13
             otp: 24.3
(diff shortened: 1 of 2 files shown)


Argonus:
@joshuawscott 
I've an idea how to split migration, tests some parts of it already on local env
Could you merge
- https://github.com/kafkaex/kafka_ex/pull/475
- https://github.com/kafkaex/kafka_ex/pull/478
Both are accepted.

PR #483, opened 2023-05-09
[Kayrock Preparations] Structs and unit tests

Argonus:
## Overview
- Nest structs under struct dir to make code easier to read. This will bound all `same kind of modules` under same path in application code.
- Add more unit tests for structs, and basic functions. More will come with next PR's. I'm aiming to separate integration tests & unit tests in 100% and make it easy to do plain TDD approach.
- Add Hammox & first behaviours to be able to type-safe mocks in unit tests.
- A bit of refactor for test helper modules

## What Next?
- Describe groups endpoint using Kayrock only
- Migrate next endpoints to Kayrock (2 probably)
- Add more unit tests

## Goal
- Migrate all endpoints to Kayrock
- Refactor Unit Tests, to make them capable to helping in TDD


test/kafka_ex/network_client_test.exs
```diff
@@ -0,0 +1,68 @@
+defmodule KafkaEx.NetworkClientTest do
+  use ExUnit.Case, async: true
+
+  use Hammox.Protect,
```
joshuawscott, line 4: I've not used Hammox, but reading through and doing some testing myself, it seems like this call just creates wrappers that will additionally fail if dialyzer types don't match? In testing locally, the function calls here are definitely still calling through to the originals, so there are no real mocks created.

Argonus: I'm using Hammox quite a lot.
You have two modes there, one with `use Hamox.Protect` it will wrap function & check in in run time same way as dialyzer in compile time. It's quite useful as it allows you to validate behaviors when dialyzer sometimes is too optimistic.

Argonus:
@joshuawscott is it ok?

lib/kafka_ex/new/structs/broker.ex
joshuawscott, line 1: Do we need to move the files here? Generally the path should match the module name.

Argonus: @joshuawscott fixed.
My idea behind moving this, is to make better separation of context. Entities like this are quite often noise in many libs once they are in same dir as core API modules etc. 
I've fixed module names to contains `Structs` either

PR #484, opened 2023-05-29
Add telemetry for offset lag

LordSarevok:
### What is the goal? how is it being implemented?

Open a PR to gather feedback about a feature that our team needs.

We are interested in the metrics from the point of view of the application about the status of the offsets of a topic and partition.

Our motivation is to have apart from visibility, alerts on this metric.

These metrics can be gathered from the broker, but there is no way to extend them with proper application context information, and for our use case it would be nice.

Please, I didn't write the test yet or the readme because I'm not sure if you will accept PRs like this (I saw no telemetry in the project, and maybe is there a reason for it) and also the approach itself, maybe due my lack of knowledge on the library I wrote the code in the wrong place.

Thanks!

### DIFF
--- a/lib/kafka_ex/gen_consumer.ex
+++ b/lib/kafka_ex/gen_consumer.ex
@@ -739,7 +739,7 @@ defmodule KafkaEx.GenConsumer do
                response = %{error_code: error_code, partition: _partition}
              ]
            }
-         ],
+         ] = fetched_responses,
          state
        ) do
     state =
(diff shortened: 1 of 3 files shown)


Argonus:
Hi, hmmm that looks interesting, but I'd like to see it as optional support. Not everyone is using telemetry, and most other packages I've seen is making such a think as optional configurable.

LordSarevok:
Maybe it was easier than I thought, Do you mean using `optional: true` on the dependency definition and relying on `Code.ensure_compiled?`?

Again, thank you very much 🙇 !!

Argonus:
Sorry, I've been off for two weeks. 
Hmmm, for example here we are using it for snappy vs snappyer. Also, sth simillary is in https://github.com/cabol/nebulex.
Its perfectly fine for `optional: true`

PR #490, opened 2023-12-04
Fix GCC 13 compilation failure

VitorTrin:
`snappier` had problems with gcc 13 that were fixed in the 1.2.9. This pr bumps it's version to fix that


VitorTrin:
also, `ssl_verify_fun` was having problems, but going deep through the deps tree I found that bumping `excoveralls` solved that

Argonus (approved):

PR #496, opened 2025-07-10
Fix socket error for OTP 28

### LINKED ISSUE #495: Incompatibility with OTP 28
When using OTP 28 we get an error in socket.ex:113:

This pattern match fails:
`{:sslsocket, {:gen_tcp, port, _, _}, _} = socket.socket`

It seems that the internal structure of the sslsocket has changed with OTP 28.
This works:
`{:sslsocket, port, _, _, :gen_tcp, _, _, _} = socket.socket` 

Best regards
André

andre-tebart:
This fixes #495


bjhaid:
Thanks

PR #546, opened 2026-06-18
Fix consumer-group sync recovery

### LINKED ISSUE #545: Recoverable SyncGroup/Heartbeat error during a rebalance crashes the Manager
A recoverable SyncGroup error (e.g. `:unknown` / `:timeout` from a coordinator under load) crashes the consumer-group Manager. Because the consumer-group supervisor is strategy: `:one_for_all`, and `max_restarts: 0`, that crash tears down the entire consumer-group subtree (client + consumers) and rebuilds it from scratch — a rejoin-from-zero that re-triggers a group-wide rebalance. During the membership churn of a routine rolling deployment, this snowballs into a prolonged (~30 min observed in one of our consumers) consumption stall instead of a quick rebalance.

### Root cause

  1. The client retries `:rebalance_in_progress`. `sync_group` and` heartbeat` inherit the `always_retryable` default in the client retry loop, so the broker's normal "rebalance in progress" signal is retried `@retry_count` (3×) with no backoff. The retries are pointless (a rebalance can't complete between sub-ms retries).
  2. `Manager.sync/2` crashes on any non-`:rebalance_in_progress` error. It handles only `{:error, :rebalance_in_progress}`; everything else (including `:unknown` / `:timeout`) raises `KafkaEx.SyncGroupError`. Under `one_for_all, max_restarts: 0`, that crash is not restarted locally — the whole subtree terminates and the parent rebuilds it (new client, new sockets, fresh JoinGroup with empty member_id), which forces a group-wide rebalance. `Manager.join/2` already retries recoverable errors (incl. `:unknown`) with backoff; `sync/2` does not.

### Logs

```
  Request Kayrock.Heartbeat.V3.Request failed with error :rebalance_in_progress   (×3)
  Receiving data from broker "..." timed out
  ** (KafkaEx.SyncGroupError) Error syncing consumer group "...": :unknown
  GenServer ... terminating   (Manager.sync/2)
```

Observed on kafka_ex 1.0.0, kayrock 1.0.0; Kafka 3.7 AWS MSK

nicolazar:
This PR fixes #545 

It introduces two fixes as follows:

-  ensures that the Managers recovers instead of crashing, bounded: before `sync/2` raised `SyncGroupError` on any non-rebalance error (including `:unknown` / `:timeout` from a coordinator under rebalance churn), and the consumer-group supervisor (one_for_all, max_restarts: 0) escalated that crash into a full-subtree teardown, re-triggering a group-wide rebalance. A recoverable sync error now routes to `rebalance` (in-place rejoin), bounded at @max_sync_retries consecutive failures (reset on a successful sync) before giving up. This mirrors join API `JoinGroupRetriesExhausted`.

- stop blind-retrying the rejoin signal: `sync_group` / heartbeat inherited the RequestContext `always_retryable` default, so a normal `:rebalance_in_progress` was retried 3 times. This is pointless, a rebalance won't complete in between sub-ms retries. They now set `sync_group_retryable?` / `heartbeat_retryable?` predicates so `:rebalance_in_progress` (and, for heartbeat, `:unknown_member_id`) bubbles to the manager's existing rejoin.

New tests added: predicate unit tests; a deterministic give-up bound test (drives a real Manager against MockClient via a small test-only :client injection seam in Manager.init, inert in prod); and finally an integration test for two-member rebalance convergence.

### DIFF
--- a/lib/kafka_ex/client/client.ex
+++ b/lib/kafka_ex/client/client.ex
@@ -752,7 +752,12 @@ defmodule KafkaEx.Client do
   end
 
   defp handle_heartbeat_request(request, node_selector, state) do
-    %RequestContext{request: request, parser_fn: &ResponseParser.heartbeat_response/1, node_selector: node_selector}
+    %RequestContext{
+      request: request,
+      parser_fn: &ResponseParser.heartbeat_response/1,
+      node_selector: node_selector,
+      retryable?: &Retry.heartbeat_retryable?/1
+    }
     |> handle_request_with_retry(state)
   end
 
@@ -772,7 +777,12 @@ defmodule KafkaEx.Client do
   end
 
   defp handle_sync_group_request(request, node_selector, state) do
-    %RequestContext{request: request, parser_fn: &ResponseParser.sync_group_response/1, node_selector: node_selector}
+    %RequestContext{
+      request: request,
+      parser_fn: &ResponseParser.sync_group_response/1,
+      node_selector: node_selector,
+      retryable?: &Retry.sync_group_retryable?/1
+    }
     |> handle_request_with_retry(state)
   end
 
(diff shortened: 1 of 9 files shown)


Argonus:
🏅 like always

PR #564, opened 2026-06-30
Allow consumers to fetch from closest replica (KIP-392 a.k.a. "rack-awareness")

soundmonster:
[KIP-392](https://cwiki.apache.org/confluence/spaces/KAFKA/pages/95653762/KIP-392+Allow+consumers+to+fetch+from+closest+replica) allows follower fetching.
Consumers can fetch from a node that is not the _leader_ for a given partition, but holds a replica.
This is important as it allows to consume from brokers within the same availability zone / data center, saving transfer costs.

For this to work, a consumer must:
* advertise its `rack_id` to the brokers, and
* follow hints from brokers as to where to fetch a given topic/partition from (`preferred_read_replica`)

This PR does exactly this:
* forward `client_rack` option for GenConsumer to brokers on fetch
    * building on top of existing `rack_id` support for fetch requests
    * building on top of existing `preferred_read_replica` support for fetch responses
* keep track of preferred read replicas in cluster metadata, and use them to pick brokers for fetch requests
* add brokers' rack IDs to telemetry metadata - this should allow users to monitor what racks KafkaEx is consuming from

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -300,3 +300,61 @@ jobs:
       - name: Retry failed tests (2nd retry)
         if: steps.retry-1.outcome == 'failure'
         run: mix test --only auth --failed
+
+  rack-awareness-tests:
+    name: Rack Awareness Tests
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+
+    env:
+      MIX_ENV: test
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Elixir & Erlang
+        uses: erlef/setup-beam@v1
+        with:
+          elixir-version: "1.18"
+          otp-version: "27.2"
+
+      - name: Restore dependencies cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            deps
+            _build
+          key: ${{ runner.os }}-mix-27.2-1.18-${{ hashFiles('**/mix.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-mix-27.2-1.18-
+
+      - name: Install dependencies
+        run: |
+          mix local.rebar --force
+          mix local.hex --force
+          mix deps.get
+          mix deps.compile
+
+      - name: Check Docker availability
+        run: |
+          docker --version
+          docker compose version
+
+      - name: Docker-compose up
+        run: ./scripts/docker_up.sh
+
+      - name: Run Rack Awareness Tests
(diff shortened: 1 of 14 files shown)


soundmonster:
One side effect of this PR is that even if neither the client nor the broker support / intend to maintain rack-awareness, KafkaEx will follow broker hints as to which node to send the next fetch request to. I'm fairly confident that this won't be a problem as this is how the protocol is intended to work, yet I'm not sure if we should put this behavior behind a feature switch.

I would like to hear the maintainers' opinions about this - what is your stance on feature switches for client behaviour? Is there any historic precedent? 

Should we only follow the preferred replica hints when a client rack is specified? I considered doing this right from the start, yet these are two separate concerns: 1) telling the broker what rack the client is in, and 2) fetching from preferred replica instead of partition leader. 1 doesn't make sense without also having 2, but 2 can be valid and desired in its own right.

Argonus:
Hey @soundmonster, sorry it took me a while, but i had to jump to source code.

If i understand this correctly, i see one blocker from merghing.

The read-replica hint leaks into Produce. Select_node(:topic_partition) is shared by fetch and produce/list_offsets. With pick_preferred_or_leader inside it and the cache keyed by {topic, partition}, a client that produces and consumes the same partition will route the produce to a follower → not_leader_or_follower. It's hidden because merge_brokers/2 drops preferred_replicas on every metadata refresh, which will happen on errors. 

My suggestion is to make this `fetch` only.

Let me know if you'll pick it or i shall continue from what you setup. 👋

soundmonster:
Hey @Argonus, thank you so much for looking into it! I was so focused on the consumer side that I totally missed the producer, really good catch on your side! I can get back to this some time after mid next week, when I have access to a computer again 🏖️

soundmonster:
@Argonus here's my proposed solution (see my [last commit]( https://github.com/kafkaex/kafka_ex/pull/564/commits/97e1059a13b95ea714f8ebadc8206f63be2041ef)): 
* introduce a new node selection strategy `:topic_partition_replica` that permits replicas
* revert behavior of the `:topic_partition` strategy to only return leaders, as before this PR
* add two tests to verify that writes go to partitiion _leaders_ even if a preferred replica exists and is distinct from the leader (a regression test for the behavior that you've outlined).

Argonus:
👁️

soundmonster:
I see there are conflicts between this branch and `master`, I would resolve them with a rebase but don't want to interrupt your review. Please let me know when it's a good time for a rebase.

